PR #52, opened 2022-02-04
Generate includes from PHP source directly

iggyvolz:
This PR generates the `.h` files directly from the PHP source code (committing them to the docker so as not to require a C compiler at runtime), using srcml to extract the relevant symbols.  This adds support for PHP 7.4 and 8.1, as well as thread-safe versions (although 7.4 needs some work to get tests passing since the code assumes PHP 8 functionality - not sure if it's worth bothering).


include/Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+RUN php process.php linux
+FROM scratch AS php-8-1-x64-zts-linux
+COPY --from=php-8-1-x64-zts-linux-build /engine-8-1-x64-zts-linux.h /
+COPY --from=php-8-1-x64-zts-linux-build /constants-8-1-x64-zts-linux.php /
+
+FROM php:8.1.0-bullseye AS php-8-1-x64-nts-linux-build
+RUN apt update
+RUN apt install libffi-dev
+RUN docker-php-ext-install ffi
+RUN curl http://131.123.42.38/lmcrs/v1.0.0/srcml_1.0.0-1_ubuntu20.04.deb > srcml.deb
```
iggyvolz, line 18: Unfortunately the official srcML download link is HTTP - https://www.srcml.org/ - not an issue at runtime unless you're regenerating the headers but it's definitely a security issue.

src/Reflection/ReflectionExtension.php
```diff
@@ -109,7 +109,11 @@ public function getGlobalsSize(): int
      */
     public function getGlobals(): ?CData
     {
-        return $this->moduleEntry->globals_ptr;
+        if(ZEND_THREAD_SAFE) {
```
iggyvolz, line 112: https://github.com/php/php-src/blob/1c33ddb5e5598c5385c4c965992c6e031fd00dd6/Zend/zend_modules.h#L87 - not sure how to deal with this in the thread-safe case

src/System/Compiler.php
```diff
@@ -27,65 +28,65 @@ class Compiler
...
-    public const COMPILE_DELAYED_BINDING = (1 << 5);
+    public const COMPILE_DELAYED_BINDING = Defines::ZEND_COMPILE_DELAYED_BINDING;
 
     /* disable constant substitution at compile-time */
-    public const COMPILE_NO_CONSTANT_SUBSTITUTION = (1 << 6);
+    public const COMPILE_NO_CONSTANT_SUBSTITUTION = Defines::ZEND_COMPILE_NO_CONSTANT_SUBSTITUTION;
 
     /* disable usage of builtin instruction for strlen() */
-    public const COMPILE_NO_BUILTIN_STRLEN = (1 << 7);
+//    public const COMPILE_NO_BUILTIN_STRLEN = Defines::ZEND_COMPILE_NO_BUILTIN_STRLEN;
```
iggyvolz, line 53: Looks like this symbol was removed in PHP 8.1 - unused in the rest of the repo

src/Reflection/ReflectionValue.php
```diff
@@ -74,40 +75,40 @@ class ReflectionValue implements ReferenceCountedInterface
...
      */
-    public const IS_CALLABLE       = 12;
-    public const IS_ITERABLE       = 13;
-    public const IS_VOID           = 14;
-    public const IS_STATIC         = 15;
-    public const IS_MIXED          = 16;
+    public const IS_CALLABLE       = Defines::IS_CALLABLE;
+    public const IS_ITERABLE       = Defines::IS_ITERABLE;
+    public const IS_VOID           = Defines::IS_VOID;
+//    public const IS_STATIC         = Defines::IS_STATIC;
```
iggyvolz, line 98: These symbols were only added in 8.0 - removed to get 7.4 compatibility but they could be re-added if needed.

lisachenko:
Wow! @iggyvolz such an amazing work! It's so pity that I have returned to my project only now, but I really would like to review and merge this PR, just need to allocate some time to check how it works )

I'm going to try to review and test this on weekend and leave my feedback.

lisachenko:
Closing, this idea was inspiring 🚀 and now Claude uses almost the same approach for code generation.

Issue #79, opened 2026-07-31
Epic: engine feature-coverage roadmap — expose more of the Zend engine to userland


lisachenko:
Full audit of the generated engine surface (`include/8.4/linux-x64-nts/engine.h`, PHP 8.4 / linux-x64-nts) against the userland API in `src/`. Coverage = symbol declared in the header **and** reachable through a public z-engine API. This epic tracks the resulting feature tickets; each is a self-contained spec implemented independently (parallel PRs).

## Status — 2026-07-31 — ✅ EPIC COMPLETE

**All 14 tickets merged.** Wave A: 8/8 (PRs #80–#87). Wave B: 6/6 (PRs #88, #90, #91, #92, #93, #94).

## Coverage after both waves

| Area | Original | Final |
|---|---|---|
| Object handlers | 9/24 | **22/24** — dimension family, `count_elements`, `get_debug_info`, `clone_obj`, `get_iterator`, `get_method`, `get_closure`, `get_constructor`, `get_class_name`, `get_properties` all hooked; the remaining `free_obj`/`dtor_obj`/`get_gc` are documented non-goals (GC/shutdown re-entry unsafe — empirically confirmed by #73's GC finding) |
| Executor globals | ~5/50 | exceptions, errors, timeouts, symbol table, included files, constant table, `zend_clear_exception` |
| Compiler globals | partial | active class/op_array, lineno, auto-globals, fn flags |
| Opcode handlers | no lifecycle | full hook lifecycle with chaining |
| Class entry / surgery | partial | enum table; complete parent detachment + re-link; `addMethod` on wrappers; runtime trait aliases/precedences |
| op_array | opcodes+literals | arg_info, static vars, try/catch tables, live ranges |
| Module API | write-only | lifecycle callbacks, phpinfo output, deps |
| PHP 8.4 features | 0% | enums, property hooks (read), attributes (read), lazy-object introspection, `get_iterator` bridge |

## Roadmap — final

**Wave A — ✅ complete:** #65→#80, #66→#84, #67→#86, #68→#81, #69→#85, #70→#82, #71→#83, #72→#87 (all merged)

**Wave B — ✅ complete:**

| # | Ticket | PR | Outcome |
|---|---|---|---|
| 9 | #73 method/closure/constructor handlers | #94 | ✅ merged |
| 10 | #74 constant table + `zend_clear_exception` | #91 | ✅ merged |
[...]

Issue #50, opened 2021-07-03
[Bug] Do not allow to throw PHP exceptions within FFI Zend callbacks


lisachenko:
As throwing is restricted now in PHP, see https://github.com/php/php-src/pull/6366/commits/327e6c9d37dbb36c4a0097fcaabc9d58ecfe9323, I should either handle all exceptions from hooks internally with try..catch section silencing them or convert them to user fatal error with `trigger_error(...,E_USER_ERROR)`.

Maybe there is some advanced technique to install custom exception handler for this, but I'm not aware about this.

lisachenko:
Related to https://github.com/php/php-src/pull/6366

PR #88, opened 2026-07-31
feat(module): module lifecycle callbacks, phpinfo output and dependency wiring for AbstractModule

### LINKED ISSUE #75: feat(module): module lifecycle callbacks, phpinfo output, and dependency wiring for AbstractModule
Part of the engine feature-coverage roadmap (epic to be linked).

## Summary

`EngineExtension\AbstractModule` registers a real `zend_module_entry` but only fills `size/type/name/zend_api/zend_debug/zts/globals_size/globals_ptr`. The module cannot react to any lifecycle event, print `phpinfo()` output, or declare dependencies — `ModuleDependency` exists but the `deps` field is never written.

## Proposed API

```php
namespace ZEngine\EngineExtension;

interface ModuleLifecycleInterface {          // opt-in: wired only when the module class implements it
    public function moduleStartup(): void;    // MINIT
    public function moduleShutdown(): void;   // MSHUTDOWN
    public function requestStartup(): void;   // RINIT
    public function requestShutdown(): void;  // RSHUTDOWN
}
interface ModuleInfoInterface {
    public function displayInfo(): void;      // info_func; may echo phpinfo-style table (helper: printInfoTable(array $rows))
}
// ModuleInterface addition (or AbstractModule virtual):
public function getDependencies(): array;     // list<ModuleDependency>, written into zend_module_entry->deps
```

## Implementation sketch

- Modify `src/EngineExtension/AbstractModule.php` (+`ModuleInterface`); reuse the FFI-closure trampoline approach from `ExtensionConstructorHook` (`globals_ctor`) for `module_startup_func`, `module_shutdown_func`, `request_startup_func`, `request_shutdown_func`, `info_func`.
- `deps`: allocate a persistent, `NULL`-terminated `zend_module_dep[]` from the `ModuleDependency` list (`MODULE_REQUIRED/CONFLICTS/OPTIONAL` already defined).
- **Safety design (mandatory)**: callback trampolines are process-lifetime (immortal-by-design — extend the table in `docs/long-running.md`); every trampoline must check `Core::isShutdown()` and no-op after `Core::shutdown()`, since `module_shutdown_func` can run after the FFI bridge is torn down. Callbacks must not throw across the FFI boundary (#50).
- Native function registration (`zend_function_entry` array → `functions` field) and per-module INI (`ini_entry`) are **out of scope** — they need header regeneration; flag as follow-up in the epic.

## Test approach

- `#[Group('internal')]`: register a test module implementing both interfaces late in a request; assert `requestShutdown()` fires at request end (register-shutdown ordering vs `Core::shutdown()` documented and tested); assert `phpinfo()` output contains the module section; assert `deps` visible via `ReflectionExtension`.
- Leak scenario for the persistent `deps` array + trampolines (expected: immortal-by-design exemptions, explicitly listed).

## Header regeneration

None for callbacks/deps (`zend_module_entry`, `zend_module_dep` fully declared).

## Risk

Medium-high: engine calls these callbacks at times when PHP re-entry may be unsafe (MSHUTDOWN after request teardown). The `Core::isShutdown()` guard and "request-phase callbacks only are guaranteed" documentation are part of the deliverable.

## Acceptance criteria

- [ ] Lifecycle + info interfaces wired opt-in; deps written; docs updated (`docs/long-running.md` immortal table)
- [ ] `#[Group('internal')]` tests + leak scenario with documented exemptions
- [ ] phpstan/cs gates pass

lisachenko:
Fixes #75

## Summary

`EngineExtension\AbstractModule` now registers a functional extension instead of a bare globals allocator:

- **`ModuleLifecycleInterface`** (`moduleStartup` / `moduleShutdown` / `requestStartup` / `requestShutdown`) — wired opt-in by `register()` via FFI-closure trampolines into `module_startup_func` / `module_shutdown_func` / `request_startup_func` / `request_shutdown_func` (same approach as `ExtensionConstructorHook`/`globals_ctor`, one small `AbstractModuleLifecycleHook` family in `EngineExtension\Hook`).
- **`ModuleInfoInterface::getDisplayInfo(): array`** → `info_func` trampoline; `AbstractModule::printInfoTable()` renders the returned `label => value` rows into the module's phpinfo() section (text and HTML modes). Per your comment on the issue, the interface returns data and AbstractModule owns the rendering wiring, instead of a void `displayInfo()`.
- **`ModuleInterface::getModuleDependencies(): list<ModuleDependency>`** → persistent NULL-terminated `zend_module_dep[]` written into `->deps`; visible through the native `ReflectionExtension::getDependencies()` and enforced by the engine (conflicts at registration, required modules at startup). `ModuleDependency` was reworked into a working value object (the previous constructor assigned undeclared properties) with `required()`/`conflicts()`/`optional()` factories and relation validation.

## Safety design (mandatory part of the deliverable)

- Every trampoline checks `Core::isShutdown()` and no-ops after `Core::shutdown()`.
- Callbacks never throw across the FFI boundary (#50): ext/ffi fatals on escaping exceptions (`zend_error_noreturn(E_ERROR, "Throwing from FFI callbacks is not allowed")`), and a FAILURE from MINIT escalates to a fatal `E_CORE_ERROR` in `zend_startup_module_ex()`. Failures are contained, reported as `E_USER_WARNING`, and the engine always sees SUCCESS.
- The trampolines follow the standard hook lifecycle: `Core::shutdown()` restores the NULL pointers while the trampolines are still alive, keeping the existing invariant that no trampoline pointer survives into a structure that outlives the request — the module entry lives in the persistent module registry, and ext/ffi frees every callback trampoline at its own RSHUTDOWN (`FFI_G(callbacks)` teardown, verified against php-8.4.24 sources).
- **Only request-phase callbacks are guaranteed** (documented in `docs/long-running.md`, new "Module lifecycle callbacks" section + extended immortal-by-design table):
  - `moduleStartup()` — engine-driven MINIT trampoline during `startup()`;
  - `requestStartup()` — delivered directly by `startup()` (dl() parity; the engine only activates modules at request start);
  - `requestShutdown()` — delivered at request end by z-engine's own shutdown chain **immediately after `Core::shutdown()`** (user shutdown functions run in registration order; the engine's own RSHUTDOWN walk happens later, after the pointers were cleared);
  - `moduleShutdown()` — best-effort only: real MSHUTDOWN runs after the FFI bridge teardown (`zend_post_deactivate_modules()` for temporary modules, process shutdown for persistent ones).

## Correctness fixes required along the way

- **PHP 8.4 latent use-after-free**: `zend_register_module_ex()` stores the given `zend_module_entry` pointer directly in `module_registry` since 8.4 (`zend_hash_add_ptr`, no copy). The old code allocated the entry as owned request-lifetime FFI memory that died at the end of `register()`. The entry, module name, globals and dependency buffers are now persistent tracked allocations (immortal-by-design; all malloc-backed, invisible to the ZendMM leak gate by construction).
- **dl() parity**: registering a temporary module now sets `EG(full_tables_cleanup)` (new `Executor::enableFullTablesCleanup()`), exactly like `ext/standard/dl.c` — otherwise the engine never deactivates/destroys runtime-registered modules at request end (the precomputed handler lists are built at process startup only).
- Replaced the deprecated `['obj', 'Class::method']` callable form in `register()` and the implicitly nullable constructor parameter (removes two baseline-covered deprecations from this path).

## Deviations from the issue sketch (with reasons)

- `getDependencies()` → **`getModuleDependencies()`**: `AbstractModule` inherits the native `ReflectionExtension::getDependencies()`, which reads the registered deps back in the engine's `name => relationship` format (and is what the tests assert against). Overriding it with an incompatible return type would be an LSP violation flagged by PHPStan and would break reflection consumers.
- The issue sketch called the trampolines "process-lifetime / immortal". That is not physically possible: ext/ffi frees every callback trampoline at its RSHUTDOWN. Instead the trampolines follow the standard hook lifecycle (pointers cleared at `Core::shutdown()` — which is also why they're not listed as new immortal allocations), and the guaranteed `requestShutdown()` delivery moved into z-engine's shutdown chain; ordering vs `Core::shutdown()` is documented and asserted end-to-end.

## Out of scope (follow-ups in the epic)

- Native function registration (`zend_function_entry` array → `functions` field).
- Per-module INI (`ini_entry`).

Both need header regeneration and are flagged as follow-ups in the epic (#79). No header regeneration was needed here (`zend_module_entry`/`zend_module_dep` fully declared).

## Tests

- `tests/EngineExtension/AbstractModuleTest.php` (`#[Group('internal')]`, one process per test — module registration cannot be undone in-process): MINIT-via-trampoline + RINIT delivery, phpinfo(INFO_MODULES) section content via output buffering, deps via native `ReflectionExtension`, exception containment as warnings, dependency validation, and a child-process fixture asserting the full ordering (`moduleStartup` → `requestStartup` → conflict rejection → script end → `requestShutdown` with `Core::isShutdown() === true` → later user shutdown functions; `moduleShutdown` never delivered; exit code 0 through the engine's temporary-module destruction).
- `tests/Memory/scenarios/module-lifecycle.php` leak scenario with the immortal-by-design exemptions explicitly listed (persistent entry, name buffer, `zend_module_dep[]` + strings — malloc-backed, so the ZendMM gate stays silent by construction; trampolines freed by ext/ffi RSHUTDOWN).

## Evidence

**Debug build (PHP 8.4.24 NTS DEBUG, `--enable-debug --with-ffi`), internal group with process isolation:**

```
$ php-debug/bin/php -d ffi.enable=1 -d zend.assertions=1 -d assert.exception=1 \
    vendor/bin/phpunit --group internal --process-isolation
OK, but there were issues!
Tests: 81, Assertions: 314, Skipped: 2, Incomplete: 2.   (skips/incompletes pre-existing)
```

Leak scenarios alone (includes the new `module-lifecycle` scenario): `OK (12 tests, 72 assertions)`.

**Smoke script** (run under both the release 8.4.19 and debug 8.4.24 builds with `report_memleaks=1`, not committed):

```
moduleStartup(coreShutdown=false)
requestStartup(coreShutdown=false)
started=true
phpinfo-has-section=true
array ( 'standard' => 'Required', 'spl' => 'Optional', )
script-end
requestShutdown(coreShutdown=true)
late-user-shutdown(coreShutdown=true)
EXIT=0
```

**Gates:** `composer test` OK (180 tests, 2758 assertions), `composer phpstan` OK — no errors, **no baseline additions**, `composer cs:check` OK.

https://claude.ai/code/session_017fzpvLxVmyXjdbZYaCu1pC

---
_Generated by [Claude Code](https://claude.ai/code/session_017fzpvLxVmyXjdbZYaCu1pC)_

### DIFF
--- a/docs/long-running.md
+++ b/docs/long-running.md
@@ -86,13 +86,45 @@ inactive during shutdown-phase object destructors, and installing a new hook thr
   engine field in such setups — stacked chains keep intermediate `proceed()` targets from
   the previous cycle.
 
+## Module lifecycle callbacks
+
+`AbstractModule::register()` wires FFI-closure trampolines into the module entry's
+`module_startup_func` / `module_shutdown_func` / `request_startup_func` /
+`request_shutdown_func` slots when the module class implements `ModuleLifecycleInterface`,
+and into `info_func` when it implements `ModuleInfoInterface`. Two hard rules shape the
+delivery semantics:
+
+1. **The trampolines follow the standard hook lifecycle.** `Core::shutdown()` restores the
+   NULL pointers while the trampolines are still alive, so the module entry — which the
+   persistent module registry references directly since PHP 8.4 — never points into freed
+   libffi memory (ext/ffi frees every callback trampoline at its own RSHUTDOWN). Every
+   trampoline additionally checks `Core::isShutdown()` and no-ops after shutdown.
+2. **Callbacks never throw across the FFI boundary** ([#50](https://github.com/lisachenko/z-engine/issues/50)):
+   ext/ffi aborts the process on an escaping exception, and a FAILURE result from MINIT
+   escalates to a fatal `E_CORE_ERROR`. Failures are contained and reported as
+   `E_USER_WARNING`; the engine always sees SUCCESS.
+
+Consequences — **only request-phase callbacks are guaranteed**:
+
+| Callback | Delivery |
+|----------|----------|
+| `moduleStartup()` | guaranteed: the engine calls the MINIT trampoline inside `AbstractModule::startup()` |
+| `requestStartup()` | guaranteed: delivered directly by `startup()` for the current request (dl() parity — the engine only activates modules at request start) |
+| `requestShutdown()` | guaranteed: delivered at request end by z-engine's own shutdown chain, immediately **after** `Core::shutdown()` (user shutdown functions run in registration order and `Core::init()` registered first). The engine's own RSHUTDOWN walk happens later, after the trampoline pointers were cleared. Engine writes are already forbidden inside this callback. |
+| `moduleShutdown()` | best-effort only: real MSHUTDOWN runs after the FFI bridge teardown (for temporary modules in `zend_post_deactivate_modules()`, for persistent ones at process shutdown), where no PHP callback can be reached. It fires only if the engine destroys the module while the request is still alive. |
+
+Runtime-registered modules are wired for the current request only: after `Core::shutdown()`
+the entry keeps no callback pointers, so subsequent requests of the same process (FPM,
+workers) see the module without lifecycle callbacks unless it is registered again.
+
 ## Immortal-by-design allocations
 
 The debug-build leak gate treats the following as expected, by design:
 
 | Allocation | Why it stays |
 |------------|--------------|
-| Module entries, module name/globals buffers (`AbstractModule`) | the engine module registry references them for the process lifetime |
+| Module entries, module name/globals buffers (`AbstractModule`) | since PHP 8.4 the engine module registry stores the registered `zend_module_entry` pointer directly (no copy), so the entry and its buffers must live for the process lifetime; all are malloc-backed |
+| Persistent `zend_module_dep[]` arrays and their name/rel/version strings (`AbstractModule::getModuleDependencies()`) | referenced from the registered module entry's `deps` field for the process lifetime; malloc-backed, bounded to one array per registered module |
 | One `zend_object_handlers` block per hooked class entry | objects still dereference `->handlers` after user shutdown functions ran, so freeing at shutdown would be a use-after-free; blocks are malloc-backed, keyed by class entry address, and bounded |
 | One live libffi trampoline per installed hook | owned by ext/ffi, freed by its RSHUTDOWN |
 | Closures immortalized by `ReflectionClass::addMethod()` | the method table references the closure body for the rest of the request |
--- a/src/EngineExtension/AbstractModule.php
+++ b/src/EngineExtension/AbstractModule.php
@@ -15,9 +15,30 @@ namespace ZEngine\EngineExtension;
 
 use FFI\CData;
 use ZEngine\Core;
+use ZEngine\EngineExtension\Hook\AbstractModuleLifecycleHook;
 use ZEngine\EngineExtension\Hook\ExtensionConstructorHook;
+use ZEngine\EngineExtension\Hook\ModuleInfoHook;
+use ZEngine\EngineExtension\Hook\ModuleShutdownHook;
+use ZEngine\EngineExtension\Hook\ModuleStartupHook;
+use ZEngine\EngineExtension\Hook\RequestShutdownHook;
+use ZEngine\EngineExtension\Hook\RequestStartupHook;
 use ZEngine\Reflection\ReflectionExtension;
 
+/**
+ * Base class for userland PHP extensions (engine modules)
+ *
+ * Memory model (docs/long-running.md): since PHP 8.4 zend_register_module_ex() stores the
+ * given zend_module_entry pointer directly in the module registry - it no longer copies the
+ * structure. The entry and every buffer it references (module name, globals, dependency
+ * array and its strings) are therefore allocated persistently and stay alive for the whole
+ * process: immortal-by-design allocations.
+ *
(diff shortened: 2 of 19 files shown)


lisachenko:
Resolve conflicts now please/rebase

lisachenko:
Conflicts resolved — master (`6e51721`, including #89) merged in as `5dff924`. The only conflicting file was `AbstractModule.php`, and the overlap was semantic: #89 and this PR independently fixed the same PHP 8.4 module-entry use-after-free with the identical persistent allocation, so master's side was kept for everything it covers — including `makeRegistryKeyPersistent()` and the registry-key fix, which now runs in the persistent branch of `register()`.

One deliberate divergence from #89 kept from this branch: the module **name/globals/deps buffers** stay persistent (`newPersistentString()`/tracked persistent) rather than #89's `Core::new(..., persistent: targetPersistent())`, because for *temporary* modules that allocates ZendMM request memory — `entry->name` would dangle when the engine destroys the module after request shutdown, and the debug leak gate flags it. `Executor::enableFullTablesCleanup()` also survives: #89's request-safety work covers the registry key and ObjectStore recycling but nothing on master triggers the engine's full-table walk, which temporary modules need to be destroyed at request end (`dl()` parity).

Post-merge gates: release suite 189 tests OK; PHPStan 0 errors (no baseline additions beyond master's own); cs clean; internal group under a PHP 8.4.24 NTS debug build 83 tests / 0 errors with all 14 leak scenarios (including this PR's `module-lifecycle` and both of #89's) green; the child-process ordering fixture still passes.

---

Issue #77, opened 2026-07-31
feat: lazy objects introspection + custom engine-level iterators (requires header regen)


lisachenko:
Part of the engine feature-coverage roadmap (epic to be linked).

## Summary

Two related object-runtime features, both declared but unreachable:

1. **Lazy objects (PHP 8.4)**: `EG(lazy_objects_store)` / `zend_lazy_objects_store` are in the header; no API to detect or create lazy objects at engine level (native `ReflectionClass::newLazyGhost()` exists, but engine access enables inspecting the store and lazy-state flags of arbitrary objects).
2. **Custom iterators**: `zend_object_iterator`, `zend_object_iterator_funcs`, `zend_class_iterator_funcs` and `ce->get_iterator` are declared and never referenced — no way to give a class a native iterator (the mechanism behind `Generator`, `ArrayIterator`) without implementing userland `Iterator` with its per-step method-call overhead.

## Proposed API

```php
// ZEngine\Type\ObjectEntry
public function isLazy(): bool;                    // GC flags IS_OBJ_LAZY_UNINITIALIZED/LAZY_PROXY
public function isLazyInitialized(): bool;

// ZEngine\ClassExtension
interface ObjectGetIteratorInterface {
    public static function __getIterator(GetIteratorHook $hook): \Iterator; // engine drives a native zend_object_iterator bridged to the returned Iterator
}
// ZEngine\ClassExtension\Hook\GetIteratorHook — binds ce->get_iterator; allocates a persistent
// zend_object_iterator_funcs vtable of FFI closures (dtor/valid/get_current_data/get_current_key/move_forward/rewind)
```

## Implementation sketch

- **Regen needed**: `zend_object_iterator`, `zend_object_iterator_funcs` full definitions and lazy-object flag macros/functions (e.g. `zend_object_is_lazy` inline logic → flags constants) in `tools/generator/symbols.php`; `composer gen-headers` + commit.
- `get_iterator` lives on `zend_class_entry` (like `create_object`/`interface_gets_implemented`), so the hook follows the `CreateObjectHook` pattern, not `zend_object_handlers`.
- The iterator vtable closures are process-lifetime (immortal-by-design, document in `docs/long-running.md`); per-iteration state carried in the `zend_object_iterator.data` zval.

## Test approach

[...]

Issue #76, opened 2026-07-31
feat(reflection): PHP 8.4 property hooks + attributes at engine level (requires header regen)


lisachenko:
Part of the engine feature-coverage roadmap (epic to be linked).

## Summary

Two PHP 8.4-era metadata surfaces are declared in the headers but unreachable from the API:

1. **Property hooks**: `zend_property_info.hooks`, `zend_property_hook_kind`, `ce->num_hooked_props` are in `engine.h` but `ReflectionProperty` cannot see or manipulate hooks.
2. **Attributes**: half-built — `ReflectionClass::$attributesTable` is assigned (`src/Reflection/ReflectionClass.php:989`) but **never read**; `zend_property_info.attributes`, `zend_class_constant.attributes`, `op_array.attributes` are equally unreachable. Native `ReflectionAttribute` exists in PHP, but engine-level access enables mutation (adding/removing attributes at runtime) and raw AST-argument inspection.

## Proposed API

```php
// ZEngine\Reflection\ReflectionProperty
public function hasHooks(): bool;
public function getHook(int $kind): ?ReflectionMethod;   // ZEND_PROPERTY_HOOK_GET / ZEND_PROPERTY_HOOK_SET
public function setHook(int $kind, \Closure $body): ReflectionMethod;  // stretch goal — same machinery as ReflectionClass::addMethod()
public function removeHook(int $kind): void;             // stretch goal

// ZEngine\Reflection\ReflectionClass
public function getAttributesTable(): ?HashTable;        // finish the dead private property
// ZEngine\Reflection\ReflectionAttributeEntry (new, wraps zend_attribute)
final class ReflectionAttributeEntry {
    public function getName(): string;
    public function getLoweredName(): string;
    public function getTarget(): int;                    // offset/flags
    public function getArguments(): array;               // zend_attribute_arg list -> ReflectionValue
}
```

## Implementation sketch

- **Regen needed**: add to `tools/generator/symbols.php`: `zend_attribute` (+ `zend_attribute_arg`) under types + `layout_structs`; `zend_property_hook_kind` under enums (generates `ZEND_PROPERTY_HOOK_GET/SET` constants); run `composer gen-headers` (Docker), commit artifacts.
- `hooks` is a `zend_function**` array indexed by hook kind — wrap entries with `ReflectionMethod::fromCData()`.
[...]

Issue #68, opened 2026-07-31
feat(system): Executor state API — exceptions, error reporting, timeouts, global symbol table


lisachenko:
Part of the engine feature-coverage roadmap (epic to be linked).

## Summary

`src/System/Executor.php` exposes only 5 of ~50 `zend_executor_globals` fields (`class_table`, `function_table`, `objects_store`, `current_execute_data`, `fake_scope`). Add read-mostly accessors for the engine's error/exception/runtime state — all fields are already declared in `engine.h:747-831`.

## Proposed API (on `ZEngine\System\Executor`)

```php
// Exceptions
public function hasException(): bool;                    // EG(exception) != NULL
public function getCurrentException(): ?\Throwable;      // ObjectEntry::fromCData(EG(exception)), guarded
// (suppressing/clearing the exception needs zend_clear_exception — deferred to the constant-table ticket, requires header regen)

// Errors
public function getErrorReporting(): int;                // EG(error_reporting)
public function setErrorReporting(int $level): int;      // returns previous level
public function getUserErrorHandler(): ?callable;        // ReflectionValue on EG(user_error_handler), null if IS_UNDEF
public function getUserExceptionHandler(): ?callable;

// Runtime state (read-only)
public function getExitStatus(): int;                    // EG(exit_status)
public function getPrecision(): int;                     // EG(precision)
public function getTimeoutSeconds(): int;                // EG(timeout_seconds)
public function isTimedOut(): bool;                      // zend_atomic_bool EG(timed_out), read-only
public function getGlobalSymbolTable(): HashTable;       // EG(symbol_table) — the true global scope
public function getIncludedFiles(): HashTable;           // EG(included_files)
```

## Implementation sketch

- Modify only `src/System/Executor.php`; wrap raw fields with existing `Type\HashTable`, `Type\ObjectEntry`, `Reflection\ReflectionValue`.
- `getCurrentException()`: `EG(exception)` is a `zend_object*`; use the guarded `ObjectEntry` path (same dangling-pointer guard as `ObjectEntry::weakFor()`) and return the PHP `Throwable` instance.
- `EG(symbol_table)` is embedded (`zend_symbol_table`/`zend_array` value, not pointer) — take its address for the `HashTable` wrapper; document it as borrowed (no ownership), per `docs/long-running.md` conventions.
- No engine functions needed — pure struct field access.

## Test approach

[...]

PR #91, opened 2026-07-31
feat(system): engine constant table API + exception clearing (regenerates headers)

### LINKED ISSUE #74: feat(system): engine constant table API + exception clearing (requires header regen)
Part of the engine feature-coverage roadmap (epic to be linked).

## Summary

Two related additions that both need `tools/generator/symbols.php` entries + `composer gen-headers` (Docker):

1. **Constant table**: `EG(zend_constants)` is declared but there is no userland API to inspect, define, or undefine engine-level constants (including persistent ones and true case handling). Complements the existing `Executor::$classTable`/`$functionTable`.
2. **Exception clearing**: #68 adds `Executor::getCurrentException()`; clearing it safely requires `zend_clear_exception()` (raw `EG(exception) = NULL` leaks the object and desyncs `opline_before_exception`).

## Proposed API

```php
// ZEngine\System\Executor
public HashTable $constantTable;                       // EG(zend_constants), initialized in constructor like $classTable
public function suppressCurrentException(): void;      // zend_clear_exception(); no-op when none

// ZEngine\Reflection\ReflectionConstant (new, wraps zend_constant)
final class ReflectionConstant {
    public static function fromCData(CData $constant): self;
    public function getName(): string;
    public function getReflectionValue(): ReflectionValue;
    public function isPersistent(): bool;              // ZEND_CONSTANT_FLAGS from flags in value.u2
    public function getModuleNumber(): int;
    public function remove(): bool;                    // zend_hash_del from EG(zend_constants); user constants only
}
```

## Implementation sketch

- `symbols.php` additions: `zend_constant` under types + `layout_structs`; functions `zend_clear_exception` (and optionally `zend_register_constant`); regenerate `include/8.4/linux-x64-nts/*` via `composer gen-headers`, commit regenerated artifacts (never hand-edit — AGENTS.md).
- New `src/Reflection/ReflectionConstant.php`; modify `src/System/Executor.php`.
- Note: in PHP 8.4 the constant flags/module number live in the zval `u1.type_flags`/`u2` slots of `zend_constant.value` — mirror the engine's `ZEND_CONSTANT_*` macros; add the needed macros to the generated `constants.php` via the generator manifest.

## Test approach

- Default suite: `$constantTable->find('PHP_VERSION')` works; a `define()`d constant appears and `ReflectionConstant` reads its value/flags; `remove()` makes `defined()` false.
- `#[Group('internal')]`: `suppressCurrentException()` inside a hooked throw context; leak scenario proving the cleared exception object is released.
- `tests/EngineConstantsTest.php` / `EngineLayoutTest.php` must stay green after regeneration.

## Header regeneration

**Yes** — `zend_constant` type + layout entry, `zend_clear_exception` function (Docker-based generator).

## Risk

Medium: regeneration must not drift other layouts (layout tests gate this); constant removal of persistent/internal constants must be rejected.

## Dependencies

Builds on #68 (Executor state API).

## Acceptance criteria

- [ ] Regenerated headers committed, layout/constants tests green
- [ ] `$constantTable`, `ReflectionConstant`, `suppressCurrentException()` implemented + tested
- [ ] phpstan/cs gates pass

lisachenko:
Fixes #74

## Summary

Exposes the engine constant table and safe exception clearing to userland (requires header regeneration, included):

- **`Executor::$constantTable`** — public `HashTable` view over `EG(zend_constants)`, initialized in the constructor like `$classTable`/`$functionTable`. Bucket values are `IS_PTR` zvals pointing to `zend_constant` structures, keyed by case-sensitive constant name.
- **`Executor::suppressCurrentException()`** — clears an in-flight `EG(exception)` via `zend_clear_exception()` (releases the object, releases a parked `EG(prev_exception)`, restores `opline_before_exception`); no-op when no exception is pending. A raw `EG(exception) = NULL` write is deliberately not offered (leaks + VM desync).
- **`ReflectionConstant`** (new) — wraps `zend_constant`: `fromCData()`, `getName()`, `getReflectionValue()`, `getFlags()`, `isPersistent()`, `getModuleNumber()`, `remove()`. The flag/module accessors mirror the engine's `ZEND_CONSTANT_FLAGS` / `ZEND_CONSTANT_MODULE_NUMBER` macros over `zval.u2.constant_flags` (low byte = `CONST_*` flags, upper bits = module number). `remove()` deletes user constants from `EG(zend_constants)` via `zend_hash_del` (engine destructor releases value/name/struct) and **refuses persistent and internal-module constants** (returns `false`).

## Header regeneration

⚠️ **This PR regenerates include/ artifacts. Other regen PRs (#76, #77) touch the same generated files — merge regen PRs sequentially and re-run `composer gen-headers` on rebase.**

Generator manifest (`tools/generator/symbols.php`) additions, regenerated `include/8.4/linux-x64-nts/*` committed in the same commit:

- types + `layout_structs`: `zend_constant` (`{ zval value; zend_string *name; }`, size 24, offsets value=0, name=16 verified by the C probe and FFI validation)
- functions: `zend_clear_exception`, `zend_register_constant` (declared for the future constant-mutation API requested in the issue comments)
- defines: `CONST_CS`, `CONST_PERSISTENT`, `CONST_NO_FILE_CACHE`, `CONST_DEPRECATED`, `CONST_OWNED`, `CONST_RECURSIVE`, `PHP_USER_CONSTANT`
- `zend_exceptions.h` added to the preprocessed/probed header set in `emit.php`/`ProbeGenerator` (nothing includes it transitively, so `zend_clear_exception` was invisible before)

The regen diff is purely additive — no drift in any other symbol, layout or constant. After merging current master (which brought PR #89's regen), the artifacts were regenerated again on the merged manifest and came out **byte-identical** to the merge result.

## Testing notes (deviation, documented)

The spec asked for an internal-group test clearing a live exception from a hooked throw context. This turned out to be impossible from userland, consistent with #68's finding that PHP never executes while `EG(exception)` is set — and verified empirically: planting an exception object into `EG(exception)` by raw FFI write sends the VM into an infinite re-dispatch loop at the very next `CHECK_EXCEPTION`, because nothing switched the frame opline to `EG(exception_op)` (`LOAD_OPLINE` re-dispatches the saved opline forever). Exercising the live path needs a C shim that sets and clears within one native call, which z-engine deliberately does not ship. Per the issue's fallback option, the tests instead lock down:

- the guarded no-op contract from an engine-invoked opcode-handler context (the realistic caller context), with execution-state integrity asserted after the call
- direct `zend_clear_exception` linkage/marshalling (safe no-op without prev/current exception)
- idempotency, and a debug-build leak scenario churning suppress + constant define/reflect/remove 500× with `report_memleaks=1`

The rationale is documented in `ExecutorExceptionSuppressTest` and on `Executor::suppressCurrentException()`.

## Evidence

- `composer test` (release 8.4.19): 201 tests, 2882 assertions — OK (incl. `EngineConstantsTest` + `EngineLayoutTest`, which now also covers the `zend_constant` layout)
- internal group under debug PHP 8.4.24 NTS DEBUG (`--group internal --process-isolation`): 80 tests, 309 assertions — OK, incl. the new `constant-table-churn` leak scenario (leak-free)
- `composer phpstan`: 0 errors, **no new baseline entries**
- `composer cs:check`: clean
- Smoke run (not committed): `constantTable->find('PHP_VERSION')` ✓, `ReflectionConstant('PHP_VERSION')` value/persistent/module ✓, `define()`d constant visible with correct value/flags/`PHP_USER_CONSTANT` module ✓, `remove()` makes `defined()` false ✓, persistent removal refused ✓, suppress no-op ✓

Pre-existing finding (out of scope, not touched by this PR): `ExecutionDataTest::testGetThis` aborts with `zend_mm_heap corrupted` when the **default** suite runs on a debug PHP build — reproducible on clean master; CI doesn't see it because the default suite runs on release builds.

https://claude.ai/code/session_017fzpvLxVmyXjdbZYaCu1pC

---
_Generated by [Claude Code](https://claude.ai/code/session_017fzpvLxVmyXjdbZYaCu1pC)_


lisachenko:
Rebase please

PR #98, opened 2026-07-31
feat(reflection): generate trampoline-free functions and methods

lisachenko:
## What & why

z-engine can already inject custom native behaviour, but the fast, trampoline-free path was limited to *methods* (`ReflectionClass::addMethod()`). Everything else — object handlers, the internal branch of `redefine()`, every `AbstractHook` — assigns a PHP `Closure` to a C function-pointer field, which ext/ffi services through a **libffi trampoline** (two FFI boundary crossings per call; `AbstractHook`'s own docblock calls it "not efficient").

This PR makes op_array grafting a first-class, general capability: **generate genuine engine functions and methods** that dispatch through the normal Zend VM with **no per-call FFI trampoline** — as fast as any hand-declared PHP function.

Scope was deliberately kept to functions/methods. Forging a `zend_class_entry` from scratch was considered and dropped as too memory-safety-risky; generated methods attach to classes that already exist. No header/generator changes are needed — the feature reuses only engine primitives already imported.

### Honest boundary
Generated code runs at normal PHP/VM speed, **not literally C speed** — pure PHP cannot emit machine code (there is no JIT here). The win is removing the trampoline tax and the closure wrapper.

## Changes

- **`ReflectionFunction::addFunction($name, $closure)`** — publishes a closure's compiled op_array as a real global function in `EG(function_table)`. `function_exists()` reports it; calls run on the VM with zero FFI. Free-function analogue of `addMethod()`. Accepts any `\Closure`, including anonymous functions and first-class-callable references — `ClosureEntry` steals the op_array from the supplied closure zval.
- **`ReflectionFunction::fromClosureEntry()`** — low-level rebind of a closure's `zend_function` as a scope-less global function (rename, clear scope, strip `ZEND_ACC_CLOSURE`).
- **`HashTable::addFunctionEntry()`** — the shared IS_PTR publish + re-wrap path; `addFunction()` and `ReflectionClass::addRawMethod()` now go through one helper instead of duplicating it.
- **`Core::registerGeneratedFunction()` + `Core::shutdown()`** — generated global functions are unpublished at shutdown with the table destructor disabled (the `pDestructor = null` technique from `ReflectionMethod::fromHookCData()`), so the engine never walks a dangling entry or double-frees the immortalized closure body.

## Memory safety

The published `zend_function` lives inside the closure object, so the closure is immortalized (refcount bump), exactly as `addMethod()` already does — a bounded, immortal-by-design allocation. New doc **`docs/memory-model.md`** diagrams the three layers (PHP zvals / FFI trampoline / native C), the two crossings a trampoline costs, and why the generated path avoids them; it also carries the lifetime/ownership cheat-sheet.

## Tests

- `ReflectionFunctionTest`: `testAddFunction` (generate → call → `redefine`), `testAddFunctionRejectsExistingName` (all under the `internal` group).
- `GeneratedFunctionBenchmarkTest` (`performance` group): asserts a generated function runs within the same order of magnitude as a hand-declared one (a trampoline-backed call would be several times slower).

## Notes on review

An earlier revision added a `NativeCompiler` that used `eval()` to compile a **source string** into a closure. Per review, `eval` was rejected and the closure-stealing the reviewer asked for is already what `addFunction`/`addMethod` do — so `NativeCompiler` and its test were removed (commit `ff6df22`) rather than reworked. No source-string compilation and no new engine symbols are introduced.

## Verification notes

Verified locally on PHP 8.4.19 (NTS) via standalone smoke runs with `ffi.enable=1`, `opcache.jit=off`, `ZENGINE_STRICT_LAYOUT_CHECK=1`: `addFunction`, `redefine` on the generated function, and the shared `addMethod` path all produce correct results and shut down cleanly (exit 0). The benchmark measured a ~1.7× ratio vs a hand-declared function (the gap is the dynamic variable-function call, not a trampoline).

The dev toolchain (PHPUnit/PHPStan) could not be installed in this sandbox (the proxy blocks GitHub package downloads), so `composer test:internal`, `phpstan`, and `cs:check` are exercised by CI. The `internal` group runs process-isolated on a debug build.

https://claude.ai/code/session_0144UTqyutnGmtss9JWNdLXc


src/System/NativeFunction/NativeCompiler.php
```diff
@@ -0,0 +1,92 @@
...
+     * Compiles a signature + body into a closure through the engine compile path
+     *
+     * @param string $signature Parameter list as written in a function declaration, eg 'int $x, int $y'
+     * @param string $body      Function body statements, eg 'return $x + $y;'
+     */
+    public static function compile(string $signature, string $body): Closure
+    {
+        // Materialize the source through the normal compiler so it becomes a real op_array; the
+        // resulting closure is then grafted into the engine table by the registration paths below
+        $closure = eval(sprintf('return static function (%s) {%s};', $signature, $body));
```
lisachenko: I don’t like eval here, it is unsafe, drop it

tests/System/NativeFunction/NativeCompilerTest.php
```diff
@@ -0,0 +1,61 @@
...
+    #[Group('internal')]
+    public function testCompileFunction(): void
+    {
+        // The generated function keeps its compiled body alive until request end
+        // (immortal-by-design, see docs/memory-model.md)
+        ini_set('report_memleaks', '0');
+
+        $functionName = 'zengine_compiled_square';
+        NativeCompiler::compileFunction($functionName, 'int $x', 'return $x * $x;');
+
```
lisachenko: I believe that NativeCompiler doing simple eval is no-go. I would like to somehow steal closure - send it as closure/anonymous, then somehow extract an op_array from it from given zval as argument

src/System/NativeFunction/NativeCompiler.php
```diff
@@ -0,0 +1,92 @@
...
+ * unpublished at Core::shutdown()).
+ *
+ * Honest boundary: the generated code runs at normal PHP/VM speed, not literally C speed - pure PHP
+ * cannot emit machine code. The win is eliminating the trampoline tax, not turning bytecode into a
+ * native zif_handler.
+ */
+final class NativeCompiler
+{
+    /**
+     * Compiles a signature + body into a closure through the engine compile path
```
lisachenko: Maybe steal op_array from given Closure? Or FCC reference?

lisachenko: Good call — done in ff6df22.

It turns out the closure-stealing you're describing is already exactly what the closure-based API does. `ReflectionFunction::addFunction()` / `ReflectionClass::addMethod()` take a `\Closure`, and `ClosureEntry` grabs the closure object from the call frame's first zval argument and grafts its embedded `zend_function` (the op_array) into the engine table — no `eval`. First-class-callable syntax produces such a `\Closure` too, so both `addFunction('twice', fn (int $x) => $x * 2)` and `addFunction('name', $userFn(...))` work.

`NativeCompiler` was only `eval` sugar over `addFunction($name, $closure)`, so I removed it and its test rather than reworking it. No source strings and no new engine symbols. (I deliberately did *not* add a native `compile_string` route — it would need exporting `compile_string` + the currently-unexported `destroy_op_array` and hand-managing a transient script op_array's lifetime, which is a large new segfault surface for nothing the closure path lacks.)

---

Issue #100, opened 2026-08-02
feat(memory): first-class persistent object-graph API — a named heap surviving request shutdown


lisachenko:
## Goal

Promote the existing persistence primitives (`PersistentObjectFactory::persistentClone`, `PersistentHashTable`, `StringEntry::persistent()/persistentInterned()`) from low-level building blocks into a **documented, supported subsystem**: a named in-process registry ("heap") that stores whole object graphs which survive request shutdown in an NTS worker process.

## Proposed API sketch

```php
$heap = PersistentHeap::global();
$heap->put('graph-key', $rootObject);   // deep persistent clone of the reachable graph
$root = $heap->get('graph-key');        // next request: re-attached, usable object graph
$heap->remove('graph-key');
$heap->stats();                          // objects, strings, bytes, per-key breakdown
```

## Technical scope

- **Deep graph clone with cycle/DAG handling** — extend the DAG-aware work from #97 to arbitrary graphs (back-edges, shared sub-objects preserved as shared, not duplicated).
- **Immortalization rules**: refcount saturation / GC exclusion for persistent zvals; ensure the collector never traverses into the persistent region (`get_gc` non-goal from #79 stays respected).
- **Supported-type matrix** (must be explicit and enforced): plain userland objects and scalars/arrays — supported; resources, closures over request state, internal objects with custom handlers, objects of classes with `free_obj`-relevant state — rejected with a typed exception at `put()` time, not corrupted at read time.
- **Re-attachment semantics at request start**: class-entry identity across requests (interaction with `opcache.preload`ed vs. re-compiled classes), behavior when the class is missing or its layout changed (`getObjectSize` check → typed exception).
- **Lifecycle integration**: wire through the existing `EngineExtension` request hooks (`RequestStartupHook`/`RequestShutdownHook`) so the heap is inert during shutdown sequences; document ordering vs. `Core::shutdown()`.
- **Eviction/teardown**: explicit `remove()` + full `PersistentHeap::destroy()`; document what is irreversibly immortal (per `docs/long-running.md` table) and keep that table updated.

## Acceptance criteria

- Request-cycle simulation test proving a graph put in "request A" is readable and fully functional after a simulated RSHUTDOWN/RINIT cycle.
- Leak-plateau test: repeated put/remove of the same graph does not grow memory unboundedly.
- Rejection tests for every unsupported type in the matrix.
[...]

PR #105, opened 2026-08-02
feat(memory): named persistent heap for object graphs surviving request shutdown

lisachenko:
Implements #100 — promotes the persistence primitives (`persistentClone`, `PersistentHashTable`, persistent interned strings) into a documented subsystem: a named in-process registry storing object graphs that survive request shutdown in an NTS worker.

## API

`PersistentHeap` with `put(string $key, object $root)`, `get(string $key): ?object`, `remove(string $key)`, `stats()`, `destroy()`. Storage is injected (dependency inversion — the heap never creates its own registry table). Bootstrap is explicit: `Core::init()` → `ExtensionManager::register(new ZEngineModule())`; the single framework-wide `EngineExtension\ZEngineModule` (engine name `zengine`, required dependency on `ffi`) hosts the anchor slot — the one address a later request can rediscover — exposes `stats()` through `phpinfo(INFO_MODULES)`, and is retrieved type-safely via `ExtensionManager::get(ZEngineModule::class)` (`@template T of AbstractModule` + `class-string<T>`). Module public APIs never return `CData` (now an AGENTS.md rule).

## Design

- **Owned vs borrowed hashtables**: `new HashTable()` creates an owned engine table; `HashTable::fromCData()` wraps borrowed ones — the framework-wide ownership convention; `PersistentHashTable` reduces to allocator hooks + interned-key/sealing/destroy specifics. `HashTable` iteration is typed `int|string` keys and yields the numeric key of hashed buckets.
- **Two-phase clone** (`PersistentGraphCloner`): phase 1 validates the entire reachable graph against the support matrix with typed exceptions and zero persistent allocation on rejection; phase 2 clones with identity maps keyed by source address — shared DAG nodes cloned once, back-edges resolve to the already-recorded clone.
- **GC safety**: clones are refcount-pinned and `GC_NOT_COLLECTABLE`; the collector never buffers or scans the persistent region (the `get_gc` non-goal stays respected).
- **Address-verified re-attachment**: on first `get()` per request, class resolved by name (`MissingClassException`), object size compared (`ClassLayoutChangedException`), and every slot verified against a per-key inventory (`GraphCorruptedException`) — stale pointers are never dereferenced; all checks run read-only before the first write.
- **Exact eviction**: per-key inventory tables list every malloc block exactly once; `remove()` frees precisely those.

## Support matrix (enforced at put() time, each rejection tested)

Supported: scalars, strings, arrays (any keys, nested), plain userland objects (std handlers), uninitialized typed slots. Rejected: closures, resources, references, internal-class objects, hooked/non-standard-handler classes, enum cases (re-attachment by case name documented as future work), lazy ghosts/proxies, dynamic properties, `USE_GUARDS` classes.

## Verification

- Request-cycle subprocess scenario: 22 ordered markers covering DAG/cycle/array fidelity, inertness, fresh store handles, GC survival, eviction, bootstrap-refusal, singleton access, phpinfo stats.
- Leak-plateau: 1200 put/get/remove cycles with <2 MiB VmRSS drift.
- `composer test`: 310 tests / 3384 assertions OK. `composer phpstan`: level max, no errors, baseline delta zero. `composer cs:check`: clean.

## Known limitations (documented in docs/persistent-heap.md)

- Same-size layout changes evade the object-size check (inherent to the issue-specified check).
- Refcounted overwrites of persistent slots are refused at next re-attachment, not at write time.
- Debug builds may report unfreed property caches for graphs read but never cycled/removed before process end; release builds unaffected.
- Two pre-existing engine-interaction defects were found and flagged (#108 double `Core::init()` CData invalidation; #109 `AbstractModule::getGlobals()` zval-type cast) — `@todo` at the bypass site.
- Composer auto-registration of userland extensions is noted as planned direction only.

Refs #100


src/Type/HashTable.php
```diff
@@ -54,6 +54,8 @@
  *     zend_long         nNextFreeElement;
  *     dtor_func_t       pDestructor;
  * };
+ *
+ * @implements IteratorAggregate<int|string|null, ReflectionValue>
```
lisachenko: key can be only int|string, null should be illegel

tests/Memory/scenarios/persistent-heap-churn.php
```diff
@@ -0,0 +1,100 @@
...
+    public string $label = '';
+
+    /** @var array<int|string, mixed> */
+    public array $items = [];
+}
+
+/**
+ * Narrows a nullable heap result (or graph edge) to a live ChurnNode
+ */
+function churnNode(?object $value): ChurnNode
```
lisachenko: why this function is ever needed? I believe we can drop this

src/Memory/PersistentHeap.php
```diff
@@ -0,0 +1,791 @@
...
+     * Returns allocation statistics: totals plus a per-key breakdown
+     *
+     * Counts are inventory blocks (objects, minted strings including metadata strings,
+     * minted array tables); bytes are the directly allocated payload bytes recorded at
+     * put() time (engine-grown table data blocks are not included).
+     *
+     * @return array{keys: int, objects: int, strings: int, arrays: int, bytes: int,
+     *               perKey: array<string, array{objects: int, strings: int, arrays: int, bytes: int}>}
+     */
+    public function stats(): array
```
lisachenko, line 267: is this information wired to the module global get information (Via phpinfo?)

src/Memory/PersistentHeap.php
```diff
@@ -0,0 +1,791 @@
...
+    }
+
+    /**
+     * Narrows a dynamically typed FFI read to CData
+     *
+     * Engine struct members read through ext/ffi carry no static type information; the
+     * generated header is the runtime ground truth, so the narrowing can never fail on
+     * a verified layout (ZENGINE_STRICT_LAYOUT_CHECK).
+     */
+    private static function asCData(mixed $value): CData
```
lisachenko, line 721: strange method, I would bette skip it, instead we can use ReflectionValue, which provides top-level PHP API around zval

src/Memory/PersistentHeap.php
```diff
@@ -0,0 +1,791 @@
...
+    {
+        assert($value instanceof CData);
+
+        return $value;
+    }
+
+    /**
+     * Narrows a dynamically typed FFI read to int (see asCData)
+     */
+    private static function asInt(mixed $value): int
```
lisachenko, line 731: better to skip - use ReflectionValue, maybe some methods or guards can be used there, eg ReflectionValue + fromValueEntry + getXXX methods

src/Memory/PersistentHeapModule.php
```diff
@@ -0,0 +1,103 @@
...
+ *     entry (and therefore the globals block) for the whole process, so the registry
+ *     address survives request shutdown even though every PHP static dies with the
+ *     request. IS_UNDEF in the slot means "no registry minted yet".
+ *  2. LIFECYCLE: requestStartup/requestShutdown forward to the heap so it is
+ *     operational exactly within the request window. Delivery guarantees and the
+ *     ordering against Core::shutdown() are documented in docs/long-running.md
+ *     ("Module lifecycle callbacks") and docs/persistent-heap.md; both callbacks are
+ *     exception-free by construction (issue #50: FFI callbacks must never throw).
+ */
+final class PersistentHeapModule extends AbstractModule implements ModuleLifecycleInterface
```
lisachenko: this module can expose some information about persistent objects in memory from the stats() method, also it can declare dependency on FFI module.

Also, I believe, we need to put everything that will be needed later for the Z-engine into one module, so my proposal is to move this part into EngineExtension namespace, give it name and declare as single-entry global module, which will incorporate information about heap, etc

This module will be one per project, it may call itself somehow. Later we will develop a mechanism for the Composer extension to declare such PHP userland extensions to register them automatically

src/Memory/PersistentHeap.php
```diff
@@ -0,0 +1,791 @@
...
+     * root registry; later requests of the same process (fresh PHP state, same engine
+     * registries) recover the registry from the module-globals anchor.
+     */
+    public static function global(): self
+    {
+        if (self::$instance !== null && !self::$instance->destroyed) {
+            return self::$instance;
+        }
+
+        $module = new PersistentHeapModule();
```
lisachenko: I don't like this silent initialization, this should be done somewhere else, maybe inside Core::init() if it will be just one single module entry per whole project. user should be able to control module initialization via DI, configs: this can be size, type, etc

lisachenko: And also would like to have indirection, make modules behaving as singletones, then user can access them. Next, all modules and extensions from all places should be registered in one place and accessed by class name, eg SomeZEngineManager->get(PersistentHeapModule::class), SomeZEngineManager->has(PersistentHeapModule::class), etc. It will use phpstan template with T class-string generic, simplifying the access.

Important rule - all module extension should not provide any public API, returning back CData, this should be incapsulated inside extension internal knowledge. Extension must provide pure PHP-native interfaces, hiding the complexity of internal world. Even better, only Z-Engine core should know about CData types, providing an API outside around ZVal/CData, which hides all the complexity of internal data structures and FFI itself together with CData. Extension may know about own internal data structures, but also tries to hide this complexity from end-users. This can be even added as important rule to AGENTS.md

src/Memory/PersistentHeap.php
```diff
@@ -0,0 +1,791 @@
...
+            $module->startup();
+        }
+
+        $anchor      = $module->anchorSlot();
+        $anchorTyped = self::asCData($anchor->u1);
+        $anchorValue = self::asCData($anchor->value);
+        if (self::asInt(self::asCData($anchorTyped->v)->type) === ReflectionValue::IS_PTR) {
+            $registry = PersistentHashTable::fromCData(Core::cast('HashTable *', self::asCData($anchorValue->ptr)));
+        } else {
+            $registry = PersistentHashTable::create();
```
lisachenko: same feedback, avoid internal creation, prefer dependency inversion principle. So, if you need a persistent hash table - it should be injected or given as an argument here.

src/Memory/PersistentHeap.php
```diff
@@ -0,0 +1,791 @@
...
+        }
+
+        $keyEntry    = StringEntry::persistentInterned($key);
+        $rawKeyValue = $keyEntry->getRawValue();
+        assert($rawKeyValue instanceof CData);
+        $stringBlocks[] = $rawKeyValue;
+
+        // All inventory tables are integer-keyed on purpose: no hidden interned-string
+        // keys are minted, so the strings inventory above stays the complete list
+        $objectsTable = PersistentHashTable::create();
```
lisachenko: just to double check - why do we use static constructor there? maybe traditional new(...)?

src/Memory/PersistentHeap.php
```diff
@@ -0,0 +1,791 @@
...
+        $classesTable = PersistentHashTable::create();
+        $sizesTable   = PersistentHashTable::create();
+        $stringsTable = PersistentHashTable::create();
+        $arraysTable  = PersistentHashTable::create();
+
+        foreach ($graph->objects as $index => $objectPointer) {
+            $classEntry = $classNamePool[$graph->classNames[$index]]->getRawValue();
+            assert($classEntry instanceof CData);
```
lisachenko: I believe it is not needed here, because getRawValue already returns CData - we need give somehow classNamePool proper list/array type

lisachenko: it will be ReflectionValue[]

src/Type/HashTable.php
lisachenko: Maybe change this constructor to actually create a hashtable and separate static fromValueXXX similarly to other classes in the framework. Then we can distinguish owned hashmaps and borrowed

lisachenko: and PersistentHashTable can be organized better then

src/Type/PersistentHashTable.php
lisachenko: this can be constructor instead, see HashTable comments

lisachenko:
All 15 review threads are addressed in 7632410..189a2a8 (3 commits). Highlights:

- **HashTable keys**: `int|string` only — and not just the annotation: `getIterator()` now yields the numeric key of hashed buckets (`bucket->h`) instead of `null`; downstream signatures aligned and the defensive null-key branches removed.
- **Ownership split**: `new HashTable()` now *creates* an owned engine table; `HashTable::fromCData()` is the borrowed wrapper — matching the framework-wide owned/borrowed convention; `PersistentHashTable` reduced to allocator hooks + interned-key/sealing specifics; all in-repo borrowed sites converted.
- **DI**: the heap never creates its storage — the registry table is injected via constructor; the silent bootstrap in `global()` is gone. Boot order is explicit: `Core::init()` → `ExtensionManager::register(new ZEngineModule())` (your DI/config point); `get()` before registration throws `ExtensionNotRegisteredException`.
- **Single framework module**: `EngineExtension\ZEngineModule` hosts the heap anchor, declares a required dependency on `ffi`, and wires `stats()` into `phpinfo(INFO_MODULES)` (the fixture asserts the heap rows appear).
- **Typed registry**: `ExtensionManager::get()/has()` with `@template T of AbstractModule` + `class-string<T>` — PHPStan resolves `get(ZEngineModule::class)` to the concrete type.
- **zval helpers dropped** in favor of `ReflectionValue::fromValueEntry()` + typed getters; the static constructor on `PersistentHashTable` replaced with `new`; the churn-scenario helper removed.
- **API purity**: no module method returns `CData` anymore, and AGENTS.md gained a "Public APIs never leak CData" rule.

Two decisions to flag for you:
1. **Module engine name is `zengine`, not `z_engine`** — `AbstractModule`'s name detection can't produce the underscore ("ZEngine" has no camelCase boundary) and the final base constructor leaves no override point. If you prefer `z_engine`, the cheapest path is making `detectModuleName()` protected (or an explicit-name constant hook) — say the word and I'll add it.
2. The class-name pool turned out to hold the interned key blocks, so it's typed `array<string, StringEntry>` rather than `ReflectionValue[]` — the redundant conversion you pointed at is gone either way (root cause was `StringEntry::getRawValue()` being nullable on a non-nullable property; narrowing it also retired five accumulated guards elsewhere).

Gates: 310 tests / 3384 assertions OK, PHPStan level max clean (zero baseline growth), cs clean. Composer auto-registration for userland extensions is noted in docs/persistent-heap.md as planned direction, per your comment.

---

Issue #51, opened 2022-02-03
Support of zend_observer


ezimuel:
@lisachenko do you think is it possible to support the new [PHP 8 observer](https://www.datadoghq.com/blog/engineering/php-8-observability-baked-right-in/) features in z-engine? I'm wondering if I can intercept PHP 8 functions and class methods call using these new internals.

lisachenko:
Hello, @ezimuel! Sorry for the late reply )) But even late rely is still better than nothing

Your mentioned feature is interesting and can be used indeed, but I don't have time for OSS now, thus almost second year without touching my projects at all...

ezimuel:
@lisachenko yes, even if late I appreciate your reply, thanks. I think that this library is a great idea and I would love to experiment using it. I'll try to do something in my (limited) spare time :-)

lisachenko:
Technical spec to take this out of "intentionally out of scope" (#79) and implement it:

## Scope

Bridge the `zend_observer` fcall API to userland: a `System/Hook/ObserverHook` with `begin`/`end` callbacks, following the same install/uninstall/reinstall lifecycle as `OpCodeHook`.

## Symbols to export (`tools/generator/symbols.php`)

- `zend_observer_fcall_register` (global registration, startup-time)
- `zend_observer_add_begin_handler` / `zend_observer_add_end_handler` (+ `zend_observer_remove_begin_handler` / `remove_end_handler`) for per-function targeting
- `zend_observer_fcall_op_array_extension` (variable — the reserved op_array extension slot)
- Regenerate headers per the AGENTS.md host recipe; the byte-identical pre-check against the committed artifacts is mandatory before applying the manifest change.

## Known constraints to resolve (design section of the PR must address each)

1. **Registration timing**: `zend_observer_fcall_register` is only legal before the engine finishes startup — for z-engine that means the `opcache.preload` path (`Core::preload()`). Document that observer support requires preload mode; `Core::init()`-only setups get a typed exception, not a silent no-op.
2. **Already-compiled op_arrays**: observer handlers are stamped into op_arrays at compile time (`ZEND_OBSERVER_ENABLED`, run_time_cache slot). Investigate retroactive stamping for functions compiled before registration; if not safely possible, document the boundary (only code compiled after registration is observed) and cover it with a test.
3. **Callback safety**: FFI callbacks must never throw (#50) — same containment strategy as the existing hooks (catch-all + `E_USER_WARNING`).
[...]

PR #106, opened 2026-08-02
feat(system): zend_observer fcall bridge with verified per-function attach path

lisachenko:
Implements #51 — bridges the engine's `zend_observer` fcall begin/end handlers to userland callbacks via `System/Hook/ObserverHook`, following the standard hook lifecycle and `Core` registry semantics.

## API

```php
Core::observeFunction(
    $function,   // zend_function* via ReflectionFunction::getRawFunctionPointer()
    fn(ExecutionData $frame) => ...,                       // begin
    fn(ExecutionData $frame, ?ReflectionValue $ret) => ..., // end (nullable — see below)
);
```

## Empirically established engine constraints (docs/observer-hook.md has the full analysis)

- **Userland can never enable the observer machinery.** `zend_observer_post_startup()` freezes the config at the tail of module startup — *before* `opcache.preload` runs. Forcing `ZEND_OBSERVER_ENABLED` retroactively corrupts op_arrays sized without observer slots (verified segfault in gdb). The hook therefore requires the preload boot path AND a startup-time (MINIT) provider, and refuses with a typed `ObserverException` otherwise — never a silent no-op, never a memory-unsafe write.
- **End handlers cannot run during exception unwinding under FFI**: ext/ffi aborts in C before any PHP executes when `EG(exception)` is set. The end callback is therefore nullable (begin-only hooks for throwing functions), and the abort behavior is pinned by a sacrificial-child test so a future ext/ffi change surfaces.
- **Lazy observer slots**: attaching to a never-called function is UB in the raw engine API; `install()` resolves `ZEND_MAP_PTR` semantics (`Core::mapPtrGet()`), warms user-function run-time caches, and primes slots with the engine's own `NOT_OBSERVED` sentinel.

## End-to-end verification

A 52-line test-only C provider (`tests/fixtures/observer-enabler/`, built on demand via phpize, cleanly skipped when the toolchain is absent) enables the machinery at MINIT while observing nothing; the firing test then asserts real begin/end events from subprocesses: simple call with return value, clean detach, nested ordering (outer-begin → inner-begin → inner-end → outer-end), exception propagation with begin-only hooks, callback-exception containment (#50), and internal-function observation (`strrev`). The fixture doubles as the documented reference "startup-time provider".

## Generator change

Adds the `zend_observer` fcall symbols and `zend_internal_run_time_cache_reserved_size` to the FFI manifest. Host regeneration per AGENTS.md; byte-identical pre-check passed before each manifest change; `engine.h` delta is purely additive; layout guard green.

## Verification

`composer test`: 265 tests / 3203 assertions OK (guard tests + preload-boundary subprocess test + firing tests). `composer phpstan`: level max, no errors, baseline untouched. `composer cs:check`: clean.

Known trade-off (documented): priming a never-called function bypasses other providers' lazy init for that function within the request — invisible with the reference enabler, an interaction to be aware of alongside a real observing extension.

Refs #51

---
_Generated by [Claude Code](https://claude.ai/code/session_01T8ixfHQ5vCoNhHYuREbm1G)_

### DIFF
--- a/docs/long-running.md
+++ b/docs/long-running.md
@@ -74,6 +74,11 @@ structures (class entries, handler blocks) never carry a pointer into the next r
 freed trampolines. After shutdown, z-engine performs no engine writes at all — hooks are
 inactive during shutdown-phase object destructors, and installing a new hook throws.
 
+`ObserverHook` (the `zend_observer` fcall bridge) follows this same lifecycle but has an extra
+boot-time constraint: it can only be installed from the `Core::preload()` path and only when a
+startup-time observer provider has enabled the engine observer machinery, otherwise it refuses with
+a typed exception rather than corrupting memory. See [observer-hook.md](observer-hook.md).
+
 ### Runtime models
 
 - **Worker loops** (RoadRunner, Swoole, ReactPHP, FrankenPHP worker mode): the whole worker
--- a/docs/observer-hook.md
+++ b/docs/observer-hook.md
@@ -12,7 +12,7 @@ $function = (new ReflectionFunction('some_function'))->getRawFunctionPointer();
 Core::observeFunction(
     $function,
     fn(ExecutionData $frame) => /* begin */,
-    fn(ExecutionData $frame, ?ReflectionValue $return) => /* end */,
+    fn(ExecutionData $frame, ?ReflectionValue $return) => /* end; omit for a begin-only hook */,
 );
 ```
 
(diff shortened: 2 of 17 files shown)


lisachenko:
Too complex and discovered restrictions makes this idea not feasible to be merged. I will keep it as a Draft for now, will decide later. Agent can stop tracking this PR, it will stay here as WIP/PoC concept only

lisachenko:
@ezimuel looks like observer API is not possible normally from userland

ezimuel:
@lisachenko I see, thanks for the tentative and the  exploration.

Issue #101, opened 2026-08-02
feat: atomic runtime function/class hot-swap API (umbrella: #64, #41)


lisachenko:
## Goal

Make repeated in-place redefinition of functions/methods/classes safe for long-running processes, and expose it as one coherent, atomic engine API. Umbrella over two existing defects plus one new capability:

- **#64** — `redefine()` leaks the previous function body. Free it properly (likely requires exporting `destroy_op_array` in `tools/generator/symbols.php`; regenerate headers per the AGENTS.md host recipe with the byte-identical pre-check).
- **#41** — `addMethod()` (and mutation generally) vs opcache: an op_array/CE living in SHM (`ZEND_ACC_IMMUTABLE`) must be detected and copied out into writable memory before mutation — and the SHM original must never be freed. Produce either a fix or an explicit, tested support matrix.
- **New: `ClassDelta` / `HotSwap` apply operation** — given freshly parsed source for an already-loaded class (via `Compiler::parseString()` + compilation), compute and apply a delta against the live `zend_class_entry`: swap changed method bodies, add new methods, update constants and default property values — **atomically**, with rollback if any step fails, and preserving any installed z-engine object handlers / hooks on the class.

## Technical scope

- **op_array lifetime**: refcount handling, `static_variables` + `ZEND_MAP_PTR` slots, `run_time_cache` reset per swap; literals and cache-slot invalidation for the swapped body.
- **SHM/immutable detection**: `ZEND_ACC_IMMUTABLE` on CE and functions, interned strings owned by opcache — copy-out rules centralized in one place instead of scattered per-API.
- **Atomicity**: stage every mutation, apply only after all validations pass; on failure restore the previous state (no half-swapped class observable).
- **Leak discipline**: the previous body is freed when safe (per #64); anything that must stay immortal is documented in the `docs/long-running.md` table.
- **Interception preservation**: swapping a method must not drop `get_method`/opcode hooks or reinstall handlers out of order.

## Acceptance criteria

- Leak-plateau test: ~1000 consecutive `redefine()`/delta-apply cycles on the same class show flat memory (closes #64).
- Opcache-enabled test run (default suite) demonstrating either working copy-out mutation or the documented rejection path (resolves #41's ambiguity).
- Delta test matrix: body change, new method, removed method (decide + document semantics), constant change, default-property change, failure→rollback case.
- `composer test`, `composer phpstan`, `composer cs:check` clean; docs page for the new API.

Issue #64, opened 2026-07-31
Free the previous function body in redefine() instead of leaking it


lisachenko:
## Context

Follow-up to the memory-lifetime overhaul (#62, merged in #63). One bounded residual was deliberately left in place and documented in `docs/long-running.md`: `FunctionLikeTrait::redefine()` on a **user-defined** function/method leaks the previous function body.

## Current behaviour

`redefine()` (`src/Reflection/FunctionLikeTrait.php`) copies the closure's `zend_function` wholesale over the original one. The original op_array's exclusive allocations — opcodes, literals, vars, live ranges — become unreachable and stay allocated:

```
Zend/zend_opcode.c(1111): Freeing ... (opcode array)
Zend/zend_compile.c(7573): Freeing ... (literals)
Zend/zend_string.h(176/252): Freeing ... (strings)
```

The leak is bounded: one previous body per redefined function, not per call in a loop over the same function... actually it *is* per `redefine()` call, so a worker that repeatedly redefines the same function grows without bound. Realistic usage (redefine once at boot) leaks a few hundred bytes per function. `ReflectionFunctionTest::testRedefine` and `ReflectionMethodTest::testRedefine` currently carry a commented `ini_set('report_memleaks', '0')` exemption for exactly this.

## Why it was not fixed in #63

A wholesale `destroy_op_array()` on the original before the overwrite is incorrect, because `redefine()` first copies the original's `common` struct into the closure's function to preserve identity — after that copy the two functions **share** raw pointers:

- `function_name` (`zend_string*`, refcounted — could be addref'd before destroy)
- `arg_info` (plain allocated array, **not** refcounted — destroying the original frees it while the redefined entry still points at it → use-after-free on any reflection of parameters)
- `doc_comment` and friends

And the two candidate primitives have export problems:

- `destroy_op_array` is a real `ZEND_API` export and could be added to `tools/generator/symbols.php`, but is too coarse while pointers are shared;
- `_efree` (for freeing just the exclusive arrays) cannot be declared safely with one signature: debug builds add `ZEND_FILE_LINE_DC` parameters that are actually used, so a hand-authored release-shape declaration would pass garbage file/line pointers on the debug CI build.

## Possible directions

1. **Unshare then destroy** — before the overwrite: addref `function_name`/`doc_comment`, duplicate `arg_info` (size is computable from `num_args` + variadic + return-type slot), then export `destroy_op_array` and destroy the original. Most complete, most surgery.
[...]

Issue #41, opened 2020-03-10
[Bug] Method ReflectionClass->addMethod is not compatible with opcache


lisachenko:
With enabled opcache, function redefinition doesn't work:

```
1) ZEngine\Reflection\ReflectionFunctionTest::testRedefine
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'Yes'
+'Test'

~/z-engine/tests/Reflection/ReflectionFunctionTest.php:84
``` 

Also, there is an assertion error:
```
~/src/php-src/Zend/zend_opcode.c:134: zend_function_dtor: Assertion `function->type == 1' failed.
```

After that PHP is crashing during request shutdown:
```
Program received signal SIGABRT, Aborted.
0x00007ffff5acc428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
54      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007ffff5acc428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff5ace02a in __GI_abort () at abort.c:89
#2  0x00007ffff5ac4bd7 in __assert_fail_base (fmt=<optimized out>, assertion=assertion@entry=0x123b8b7 "function->type == 1",
    file=file@entry=0x123b860 "~/src/php-src/Zend/zend_opcode.c", line=line@entry=134, function=function@entry=0x123c820 <__PRETTY_FUNCTION__.11912> "zend_function_dtor")
    at assert.c:92
#3  0x00007ffff5ac4c82 in __GI___assert_fail (assertion=0x123b8b7 "function->type == 1", file=0x123b860 "~/src/php-src/Zend/zend_opcode.c", line=134,
    function=0x123c820 <__PRETTY_FUNCTION__.11912> "zend_function_dtor") at assert.c:101
#4  0x00000000008e91aa in zend_function_dtor (zv=0x7ffff2a63f60) at ~/src/php-src/Zend/zend_opcode.c:134
#5  0x0000000000914c79 in zend_hash_destroy (ht=0x7ffff2bb3800) at ~/src/php-src/Zend/zend_hash.c:1541
#6  0x00000000008e9ebe in destroy_zend_class (zv=0x1811a60) at ~/src/php-src/Zend/zend_opcode.c:313
[...]

PR #107, opened 2026-08-02
feat: atomic runtime function/class hot-swap API

lisachenko:
Implements #101 (umbrella over #64 and #41) — makes repeated in-place redefinition safe for long-running processes and adds an atomic class-delta apply operation.

## Part 1 — #64: `redefine()` no longer leaks the previous body

New central swap engine (`FunctionBodySwap` + `PendingBodySwap` commit/rollback handle). `redefine()` now takes one body reference **per published bucket** (fixing a latent under-count when subclass method tables share the parent's `zend_function` — previously a shutdown use-after-free waiting to happen), installs a fresh scope-correct run-time cache per swap, duplicates donor-owned static-variable defaults, then destroys the previous body via `destroy_op_array`/`zend_destroy_static_vars`. A VM-stack walk keeps the previous body alive when the redefined entry is still executing (self-redefinition covered by tests).

**Leak plateau (1000 cycles, 16 KiB tolerance):** function/method redefine overhead over the engine's own per-eval compile baseline is **exactly 0 bytes**; pure swap churn with fixed donors grows **0 bytes**. The `report_memleaks` exemptions on the redefine tests were removed.

## Part 2 — #41 resolution: centralized SHM handling

`System/SharedMemory` + typed `SharedMemoryException`. Verified by an `opcache.enable_cli=1` subprocess matrix test:
- Immutable **global function** + `redefine()` → **copy-out supported** (per-process function-table bucket repointed to a writable container; SHM never written, never freed).
- Immutable **class** mutation (method redefine, `addMethod`, `removeMethods`, trait config, `HotSwap`) → **documented, tested rejection** (the method table lives inside the SHM class entry; no writable per-process slot exists).
- Runtime-declared symbols keep the full surface under opcache.

## Part 3 — atomic `ClassDelta`/`HotSwap`

`HotSwap::prepare(string $source)` parses and compiles a donor CE (dtor-disabled unpublish/eval/republish), diffs conservatively against the live class, then stage-commits with a full undo stack — a failure at any point rolls back byte-exactly (verified via an injection point). Delta matrix tested (18 tests): body change (propagates to linked subclasses), method add/remove, constant change/add, instance+static default changes, rollback, and preservation of installed z-engine handlers/hooks. Donors are destroyed after apply/discard, so repeated cycles amortize to ~0 growth.

## Generator change

Exports `destroy_op_array`, `zend_destroy_static_vars`, `destroy_zend_class`, `zend_array_dup`, and the `ZEND_HAS_STATIC_IN_METHODS` define (fixing a misspelled manifest entry the probe silently skipped). Byte-identical pre-check passed before the manifest change; layout guard green.

## Verification

`composer test`: 279 tests / 3265 assertions OK (incl. subprocess opcache matrix + leak-plateau tests; debug-gate churn scenarios added for CI's debug job). `composer phpstan`: level max, no errors. `composer cs:check`: clean.

## Documented limitations

- Each `prepare()` costs ~1 KiB of request-arena (donor compile, same as `eval`), not individually freeable.
- Suspended generator/fiber frames of a swapped entry are not discoverable by the stack walk — documented hazard.
- Rejected delta classes: hierarchy changes, property-surface changes, constant removal, magic/constructor add-remove, inherited-method overrides, interfaces/traits/enums.
- Pre-existing `removeMethods()` keeps its old delete semantics (left untouched to avoid scope creep — noted for follow-up).

Refs #101, Refs #64, Refs #41

---
_Generated by [Claude Code](https://claude.ai/code/session_01T8ixfHQ5vCoNhHYuREbm1G)_


src/System/SharedMemory.php
```diff
@@ -0,0 +1,139 @@
...
+            );
+        }
+
+        $writableEntry = Core::trackedNew('zend_function', true);
+        Core::memcpy($writableEntry, $functionEntry, Core::sizeof($writableEntry));
+
+        // The writable copy is not opcache-shared anymore; everything it points at
+        // still is, so the body must be replaced (not freed) by the caller
+        $commonPointer = $writableEntry->common;
+        assert($commonPointer instanceof CData);
```
lisachenko: What I don’t like here and everywhere - these assertions everywhere, that looks ugly. My proposal here is to use object shapes around all core methods that return CData/Zval/etc https://phpstan.org/writing-php-code/phpdoc-types#object-shapes do not declare as phpDoc inline variable, instead rely on return values, so it will be centralized. Record this as a small rule for agents as well

lisachenko: Also, maybe we can use special phpstan.neon metadata to declare additionally such shapes as custom class tyoes

src/System/SharedMemory.php
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * @copyright Copyright 2019, Lisachenko Alexander <lisachenko.it@gmail.com>
+ *
+ * This source file is subject to the license that is bundled
+ * with this source code in the file LICENSE.
+ *
+ */
+declare(strict_types=1);
+
+namespace ZEngine\System;
```
lisachenko: We have already Memory namespace, better to organize files there, so all memory-related classes and methods will live there (you can also move existing classes in namespace if needed for better logical placement)

src/System/SharedMemory.php
```diff
@@ -0,0 +1,139 @@
...
+        return ($classFlags & Core::ZEND_ACC_IMMUTABLE) !== 0;
+    }
+
+    /**
+     * Checks if the given zend_function entry lives in opcache shared memory
+     *
+     * Only user functions can be opcache-shared; internal functions are persistent
+     * process memory, which is a different lifetime class entirely.
+     */
+    public static function isImmutableFunctionEntry(CData $functionEntry): bool
```
lisachenko: This can be stored inside our ReflectionFunction instead, as it is related to the function

lisachenko: Propose isImmutable(): bool

src/Memory/SharedMemory.php
```diff
@@ -0,0 +1,139 @@
...
+{
+    /**
+     * This is an utility class, no instances needed
+     */
+    private function __construct() {}
+
+    /**
+     * Checks if the given zend_class_entry lives in opcache shared memory
+     */
+    public static function isImmutableClassEntry(CData $classEntry): bool
```
lisachenko: This can be stored inside our ReflectionClass instead, as it is related to the class. isImmutable(): bool

src/System/SharedMemory.php
```diff
@@ -0,0 +1,139 @@
...
+     * @param HashTable|ReflectionValue[]  $table         Function table whose bucket should be repointed
+     * @param string                       $lowerKey      Lowercased bucket key of the entry
+     *
+     * @return CData Writable zend_function pointer now published in the table
+     */
+    public static function copyOutFunctionEntry(CData $functionEntry, HashTable $table, string $lowerKey): CData
+    {
+        $bucketValue = $table->find($lowerKey);
+        if ($bucketValue === null) {
+            throw new SharedMemoryException(
```
lisachenko: Usually we prefer static method constructors for exception, to be consistent

lisachenko:
All review points addressed in eed1dd0..57a9392:

- **Object shapes as named config types**: nine aliases in `phpstan.dist.neon` `parameters.typeAliases` (`ZendFunctionShape`, `ZendOpArrayShape`, `ZendClassEntryShape`, `ZvalShape`, …), used as `@return` types on centralized accessors (new `System/EngineStructs` + `FunctionLikeTrait::getCommonPointer()`). ~100 call-site asserts deleted across the hot-swap files. One technical note: `FFI\CData` is final, so `FFI\CData&object{...}` is unresolvable — the aliases intersect `stdClass&object{...}` instead, which has a nice side effect: property-name typos on shaped values now fail PHPStan (the blanket CData ignores don't apply). The raw CData handle stays in scope for FFI hand-offs; the shaped view serves field access. Kept asserts are the genuinely unshapeable ones: offset reads (centralized into `EngineStructs::counterValue/zvalAt/cdataAt`), `Core::call()` mixed returns, dynamic `->{$field}` access, and runtime refcount/presence guards.
- **AGENTS.md**: new rule "Engine structs are typed by shape, not by call-site asserts" (naming the config aliases as the extension point), next to the "Public APIs never leak CData" rule.
- **Memory namespace**: `SharedMemory` + `SharedMemoryException` moved to `ZEngine\Memory`; nothing else was clearly misplaced, so nothing else moved.
- **isImmutable()**: `FunctionLikeTrait::isImmutable()` (covers `ReflectionFunction`/`ReflectionMethod`) and `ReflectionClass::isImmutable()`; `SharedMemory` keeps only the shared detection + copy-out mechanics.
- **Named exception constructors**: `SharedMemoryException::immutableClassMutation()/immutableMethodTable()/functionNotPublished()` replacing inline `new`, matching the codebase style.

Gates: 334 tests OK, PHPStan level max clean on a cold cache with zero baseline growth, cs clean.

---

src/System/EngineStructs.php
```diff
@@ -0,0 +1,199 @@
...
+    {
+        return $struct;
+    }
+
+    /**
+     * Views a pointer as a zend_function entry (user function union view)
+     *
+     * @return ZendFunctionShape
+     */
+    public static function functionEntry(CData $functionEntry): object
```
lisachenko: This entire class with assertions is not needed, instead, you should put all object shapes into relevant reflection classes, so they will be returning proper shapes from nested internal CData fields

src/System/ExecutionData.php
```diff
@@ -121,6 +121,23 @@ public function getFunction(): \ReflectionFunctionAbstract
         return $reflection;
     }
 
+    /**
+     * Returns the numeric address of the zend_function this frame executes
+     * (null for frames without a function entry, eg top-level pseudo frames)
+     *
+     * @internal used by FunctionBodySwap to detect entries with live frames
+     */
+    public function getFunctionEntryAddress(): ?int
```
lisachenko: It would be better to return here ReflectionFunction and to have reflectionFunction->getAddress()

src/System/EngineStructs.php
```diff
@@ -0,0 +1,199 @@
...
+        return $counterValue;
+    }
+
+    /**
+     * Reads one zval slot of an engine zval table (default properties, literals)
+     *
+     * The raw handle is returned (wrap it with zval() for field access): slot
+     * values travel into FFI primitives that require CData arguments.
+     */
+    public static function zvalAt(CData $zvalTable, int $index): CData
```
lisachenko: Not sure if it should be there or not, it looks like operation on array. Then maybe we need either to work with hashmap or create a helper around array of structures in memory and make it phpstan template generic, give it entry object shape, so phpstan will be happy to work with raw object arrays in memory

src/System/EngineStructs.php
```diff
@@ -0,0 +1,199 @@
...
+        return $shaped;
+    }
+
+    /**
+     * Reads an engine counter cell (eg the uint32_t behind op_array.refcount)
+     *
+     * Offset reads on CData cannot be expressed as an object shape, so the numeric
+     * narrowing for counter dereferences is centralized here instead.
+     */
+    public static function counterValue(CData $counterPointer): int
```
lisachenko: This should be in OpArray or opcache relevant file

src/System/EngineStructs.php
```diff
@@ -0,0 +1,199 @@
...
+
+        return $shaped;
+    }
+
+    /**
+     * Views a pointer as a zend_property_info
+     *
+     * @return ZendPropertyInfoShape
+     */
+    public static function propertyInfo(CData $propertyInfo): object
```
lisachenko: Same, put into ReflectionProperty

lisachenko:
Reorganized per the follow-up review — `EngineStructs` is deleted; the shapes now live on the classes that own each struct, and contiguous struct arrays go through one generic view.

**Accessors redistributed to owners:**
- `ZendOpArrayShape` / common pointer / `getAddress(): int` → `FunctionLikeTrait` (with the existing op_array introspection family; no separate OpArray type exists — `OpLine` is a single `zend_op`).
- `ExecutionData::getFunctionEntry(): ?ReflectionFunction` replaces the raw-address accessor; the live-frame self-execution walk compares `getAddress()`.
- `ZendClassEntryShape` + borrowed table accessors → `ReflectionClass`; `isImmutable()` reads its own shape and `SharedMemory` delegates to it.
- `ZvalShape` → `ReflectionValue::getZvalShape()`; `ZendPropertyInfoShape` → `ReflectionProperty`; `ZendClassConstantShape` → `ReflectionClassConstant` (static views — instance `fromCData()` can't wrap hot-swap donor structs). Closure/hashtable shapes dropped in favor of the existing `ClosureEntry`/`HashTable` APIs.

**Offset reads** split three ways per your guidance: class method/constant/property tables now go through the `HashTable` path (your "work with hashmap" option); contiguous in-memory struct arrays (literals, default-property/static tables, interface lists) go through a new `@template T` `Type\StructArray` with `at(int): T` typed access; the lone body-refcount counter cell stays a private `FunctionBodySwap` helper (a bare `uint32_t` is neither array-of-structs nor hashtable — it's that class's ownership-accounting domain).

Assert count in the five hot-swap files: 141 → 41 → **27**, all now genuine runtime guards (refcount floors, computed-slot presence, nullable-name invariants) or `Core::call()` mixed returns. AGENTS.md rewritten: shapes are config aliases, shaped accessors live on the owning class, struct arrays via `StructArray`, no standalone helper, no call-site re-asserts.

One technical note carried forward: `FFI\CData` is final, so aliases intersect `stdClass&object{...}`; each owning accessor does the single `@var` widening once (`asStructView()`), and property-name typos on shaped values fail PHPStan.

Gates: 334 tests OK, PHPStan level max clean (cold cache, zero baseline growth), cs clean, worker-loop soak OK.

---

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+    /**
+     * Test-only failure injection point for the apply protocol
+     *
+     * When set, the closure is invoked with each operation label (eg
+     * "method.change:greet") right before the operation executes; a throw from it
+     * exercises the rollback path.
+     *
+     * @internal
+     */
+    public static ?\Closure $applyFailureInjector = null;
```
lisachenko: let's drop this - I don't like this fault injection. Either we can test existing error/normal behavior or not. If not - remove such fake tests

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+            return;
+        }
+        $this->donorReleased = true;
+        self::destroyClassEntry($this->donorEntry);
+    }
+
+    private function assertUsable(): void
+    {
+        if ($this->isApplied) {
+            throw new HotSwapException('This class delta has already been applied');
```
lisachenko: For all exception - prefer the static method constructors in the exception class. May be recorded as a short rule for agents

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+            (self::$applyFailureInjector)($operationLabel);
+        }
+    }
+
+    /**
+     * Bounds a raw engine counter for struct-array view construction
+     *
+     * @return int<0, max>
+     */
+    private static function slotCount(int $count): int
```
lisachenko: drop such method, we don't want to much assertions

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+        '__serialize',
+        '__unserialize',
+    ];
+
+    private bool $isApplied     = false;
+    private bool $isDiscarded   = false;
+    private bool $donorReleased = false;
+
+    /**
+     * @param array<string, CData> $changedMethods       Lowercased name => donor zend_function*
```
lisachenko: what about wrapping this into ReflectionMethod?  then we will incapsulate raw access inside the ReflectionMethod, keeping all other classes clean enough without dealing too much with raw FFI/CData

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+     * @param array<string, CData> $addedMethods         Lowercased name => donor zend_function*
+     * @param array<string, ?CData> $removedMethods      Lowercased name => inherited ancestor
+     *                                                   zend_function* that becomes visible again, or null
+     * @param array<string, CData> $changedConstants     Constant name => donor zend_class_constant*
```
lisachenko: Same, maybe wrap into ReflectionClassConstants created from such CData?

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+            $this->discard();
+        }
+    }
+
+    /**
+     * Computes the delta between the live class entry and the donor entry
+     *
+     * @internal called by HotSwap::prepare(); the donor ownership moves to the delta
+     */
+    public static function fromDiff(string $className, CData $classEntry, CData $donorEntry): self
```
lisachenko: let's accept ReflectionClass entries, wrapping the CData $classEntry and CData $donorEntry

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+        $methodTable    = ReflectionClass::fromCData($classEntry)->getMethodTable();
+        foreach ($donorMethods as $lowerName => $donorFunction) {
+            if (isset($originalMethods[$lowerName])) {
+                if (self::functionBodyDiffers($originalMethods[$lowerName], $donorFunction)) {
+                    $changedMethods[$lowerName] = $donorFunction;
+                }
+                continue;
+            }
+            if ($methodTable->find($lowerName) !== null) {
+                throw new HotSwapException(
```
lisachenko: prefer to use static exception constructors, this will make code easier to read and follow

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+            $addedConstants,
+            $changedPropertySlots,
+            $changedStaticSlots,
+        );
+    }
+
+    /**
+     * @return list<string> Lowercased names of methods whose bodies will be swapped
+     */
+    public function getChangedMethods(): array
```
lisachenko, line 215: maybe return ReflectionMethods keyed by their name directly? Then we can further iterate or decide about them what to do

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+    {
+        return array_keys($this->changedMethods);
+    }
+
+    /**
+     * @return list<string> Lowercased names of methods that will be added
+     */
+    public function getAddedMethods(): array
```
lisachenko, line 223: same here - return not the names, but ReflectionMethods keyed

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+     */
+    public function getRemovedMethods(): array
+    {
+        return array_keys($this->removedMethods);
+    }
+
+    /**
+     * @return list<string> Names of constants whose values will be replaced
+     */
+    public function getChangedConstants(): array
```
lisachenko, line 239: we can return map of ReflectionClassConstants here, instead of strings/CData

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+    {
+        return array_keys($this->changedConstants);
+    }
+
+    /**
+     * @return list<string> Names of constants that will be added
+     */
+    public function getAddedConstants(): array
```
lisachenko, line 247: see comment above

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+     * @internal shared with HotSwap::prepare() error paths
+     */
+    public static function destroyClassEntry(CData $classEntry): void
+    {
+        if (Core::isShutdown()) {
+            // Engine writes are forbidden after shutdown: the unpublished entry leaks
+            // (bounded - it is unreachable and carries no trampolines)
+            return;
+        }
+        $rawClass   = StructArray::ofStructs($classEntry, 1)->rawAt(0);
```
lisachenko: it is not struct, it is pointer, otherwise it is not clear, why we operate here with array and take hard-coded elements

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+        $this->isDiscarded = true;
+        $this->releaseDonor();
+    }
+
+    /**
+     * Destroys an unpublished class entry with engine semantics
+     *
+     * @internal shared with HotSwap::prepare() error paths
+     */
+    public static function destroyClassEntry(CData $classEntry): void
```
lisachenko: better to move into the ReflectionClass, to incapsulate all API around class entry

src/Reflection/ReflectionValue.php
```diff
@@ -464,6 +464,39 @@ public function getRawValue(): CData
...
+     * The declared shape (see phpstan.dist.neon typeAliases and AGENTS.md) is the
+     * single narrowing point for raw zval field access (type byte, payload union,
+     * extra word). Keep getRawValue() for FFI hand-offs; this view serves the
+     * field access.
+     *
+     * @return ZvalShape
+     *
+     * @internal shared with the hot-swap machinery (ClassDelta/SharedMemory)
+     */
+    public function getZvalShape(): object
```
lisachenko: don't add this method, just put this ZvalShape to the getRawValue() method and merge text description

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+
+        return $count;
+    }
+
+    /**
+     * Returns the shaped zval view of a class constant value
+     *
+     * @return ZvalShape
+     */
+    private static function constantValueShape(CData $constantEntry): object
```
lisachenko: this belongs purely to the ReflectionClassConstant and we can call reflectionClassConstant->getReflectionValue()->getRawData(), which will declare "@return ZvalShape"

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+
+        return ReflectionValue::fromValueEntry(Core::addr($rawValue))->getZvalShape();
+    }
+
+    /**
+     * Returns the user methods the class itself declares, keyed by lowercased name
+     *
+     * @return array<string, CData> Lowercased method name => zend_function*
+     */
+    private static function declaredMethods(CData $classEntry): array
```
lisachenko: this should accept ReflectionClass directly

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+        $classAddress    = Core::addressOf($classEntry);
+        $methodTable     = ReflectionClass::fromCData($classEntry)->getMethodTable();
+        foreach ($methodTable as $methodName => $methodValue) {
+            assert(is_string($methodName));
+            $rawFunction    = $methodValue->getRawFunction();
+            $methodFunction = ReflectionFunction::fromCData($rawFunction);
+            if (!$methodFunction->isUserDefined()) {
+                continue;
+            }
+            $methodScope = $methodFunction->getCommonPointer()->scope;
```
lisachenko: use getDeclaringClass() from ReflectionMethod - will be easier to compare, it checks already `$this->getCommonPointer()->scope`

src/HotSwap/ClassDelta.php
lisachenko: maybe we can do/move this to the implementation of ReflectionClass->getMethods()? which will override the implementation with our low-level structures. Then we can create emulation of ReflectionMethods even for non-persistent/dynamic methods, which will be residing as structures in memory.

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+
+        return $declaredMethods;
+    }
+
+    /**
+     * Returns the constants the class itself declares, keyed by name
+     *
+     * @return array<string, CData> Constant name => zend_class_constant*
+     */
+    private static function declaredConstants(CData $classEntry): array
```
lisachenko: Same, move to the ReflectionClass, implementation will override the ReflectionClass->getReflectionConstants()/getReflectionConstant()/hasReflectionConstant() API

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+        return $declaredConstants;
+    }
+
+    /**
+     * Conservative comparison of two compiled function bodies
+     *
+     * Anything that cannot be proven identical counts as changed: a spurious swap is
+     * merely redundant work, while a missed change would keep stale code running.
+     */
+    private static function functionBodyDiffers(CData $originalFunction, CData $donorFunction): bool
```
lisachenko: maybe better to have special ReflectionMethod->equals($anotherReflectionMethod)? It will not be the same, full equal + pointer same check, but we will keep low-level logic of function/method comparison close to raw structures and incapsulate this logic

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * Conservative comparison of two class constants (value and access flags)
+     */
+    private static function classConstantDiffers(CData $originalConstant, CData $donorConstant): bool
```
lisachenko: better to move to the ReflectionClassConstant->equals($anotherReflectionClassConstant) API

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+        return self::zvalDiffers($originalValue, $donorValue);
+    }
+
+    /**
+     * Conservative scalar zval comparison: non-scalar payloads always count as different
+     *
+     * @param ZvalShape $firstValue
+     * @param ZvalShape $secondValue
+     */
+    private static function zvalDiffers(object $firstValue, object $secondValue): bool
```
lisachenko: better to move to the ReflectionValue->equals($anotherReflectionValue): bool API

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+            default:
+                // Arrays, objects and constant expressions: conservatively different
+                return true;
+        }
+    }
+
+    /**
+     * Rejects removals the class entry keeps direct function pointers for
+     */
+    private static function assertRemovableMethod(
```
lisachenko: part of this implementation can be in the ReflectionMethod->isRemovable() - it should be user-defined and have necessary checks from this method.

and maybe this method will be thin or not needed at all

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+    {
+        $changedPropertySlots = [];
+        $changedStaticSlots   = [];
+        $zvalSize             = Core::sizeof(Core::type('zval'));
+        $slotBase             = Core::type('zend_object')->getStructFieldOffset('properties_table');
+        $liveClass            = ReflectionClass::fromCData($classEntry);
+        $liveEntry            = $liveClass->getClassEntry();
+        $donorClassEntry      = ReflectionClass::fromCData($donorEntry)->getClassEntry();
+        $classAddress         = Core::addressOf($classEntry);
+        foreach ($liveClass->getPropertiesTable() as $propertyValue) {
```
lisachenko: This part can be extracted into the ReflectionClass->getProperties() API, which overrides the native one implementation.

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+                continue;
+            }
+            if (($flags & Core::ZEND_ACC_STATIC) !== 0) {
+                $originalTable = $liveEntry->default_static_members_table;
+                $donorTable    = $donorClassEntry->default_static_members_table;
+                // A declared static member guarantees both classes carry the table
+                assert($originalTable !== null && $donorTable !== null);
+                $originalView = StructArray::ofZvals($originalTable, self::slotCount($liveEntry->default_static_members_count));
+                $donorView    = StructArray::ofZvals($donorTable, self::slotCount($donorClassEntry->default_static_members_count));
+                if (self::zvalDiffers($originalView->at($offset), $donorView->at($offset))) {
```
lisachenko: this can be inside ReflectionValue->equals($anotherReflectionValue): bool

src/Type/StructArray.php
```diff
@@ -0,0 +1,141 @@
...
+ * was read from and must outlive the view. Element pointer arithmetic follows
+ * the FFI type of the base pointer, so the base must be a correctly typed CData
+ * (eg `zval *` or `zend_class_entry **`).
+ *
+ * @internal
+ *
+ * @template T of object
+ * @implements IteratorAggregate<int, T>
+ */
+final class StructArray implements Countable, IteratorAggregate
```
lisachenko: + ArrayAccess too, to support dimension read/write

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+     */
+    public function getAddedConstants(): array
+    {
+        return array_keys($this->addedConstants);
+    }
+
+    /**
+     * @return list<int> Instance default-property table slots that will be updated
+     */
+    public function getChangedPropertySlots(): array
```
lisachenko: maybe getChangedProperties() + getStaticChangedProperties() ? instead of slots + to have some API inside ReflectionClass that resolves/queries slot, based on ReflectionProperty instance?

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+                $error,
+            );
+        }
+
+        // Commit: from here on nothing can fail - release everything the class
+        // previously owned that the delta replaced
+        foreach ($pendingSwaps as $pending) {
+            $pending->commit();
+        }
+        foreach ($replacedValues as $valueSnapshot) {
```
lisachenko: need to operate on ReflectionValue instances

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+     * Replaces a zval slot with the donor's value (donor keeps its own reference)
+     *
+     * The previous value is NOT released here: its snapshot is returned and the
+     * caller releases it at commit time, so a rollback can restore it byte-exact.
+     *
+     * @param list<\Closure> $undoStack Undo action list of the running apply
+     *
+     * @return CData Snapshot of the previous slot value (a zval container)
+     */
+    private static function replaceZvalSlot(CData $targetSlot, CData $donorSlot, array &$undoStack): CData
```
lisachenko: can we have this API for StructArray->replace(int $position, mixed $value): $previousValue ?

lisachenko: all size-related machinery will be there

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+    {
+        $zvalSize = Core::sizeof(Core::type('zval'));
+        $snapshot = Core::new('zval');
+        Core::memcpy($snapshot, $targetSlot, $zvalSize);
+        Core::memcpy($targetSlot, $donorSlot, $zvalSize);
+        // The slot now holds its own reference on the (possibly shared) payload
+        Core::call('zval_add_ref', Core::addr($targetSlot));
```
lisachenko: would be nice to operate via API, check ReferenceCountedTrait API, maybe zval_add_ref should be there then

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+        $undoStack[] = static function () use ($targetSlot, $snapshot, $zvalSize): void {
+            Core::call('zval_ptr_dtor', Core::addr($targetSlot));
```
lisachenko: better to operate via ReflectionValue API instead of calling low-level zval_ptr_dtor, eg ReflectionValue::fromCData(Core::addr($targetSlot))->destroy(); or similar code

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+        );
+        Core::untrackAndFree(Core::addr($container));
+    }
+
+    /**
+     * Mints an immortal zend_class_constant container adopting the donor's constant
+     *
+     * @param CData $declaringClass Target class entry the constant will belong to
+     */
+    private static function mintConstantContainer(CData $donorConstant, CData $declaringClass): CData
```
lisachenko: should we move this to the ReflectionClassConstant + better naming

src/HotSwap/ClassDelta.php
```diff
@@ -0,0 +1,911 @@
...
+            }
+        }
+
+        return $container;
+    }
+
+    /**
+     * Releases everything an added-constant container took (rollback path only)
+     */
+    private static function releaseConstantContainer(CData $container): void
```
lisachenko: this can be inside ReflectionClassConstant() as well

src/HotSwap/HotSwap.php
```diff
@@ -0,0 +1,348 @@
...
+        string $sourceCode,
+    ): CData {
+        $classTable = Core::$executor->classTable;
+
+        self::unpublishClassEntry($lowerName);
+        $donorEntry   = null;
+        $compileError = null;
+        try {
+            try {
+                eval($sourceCode);
```
lisachenko, line 190: this looks dangerous, is it possible to avoid eval? it can execute any malicious code, whereas we would like to have safe mechanism, so I'm looking into parsing source/AST/compiling new class entries in tables/etc

lisachenko: you can decide to put this as separate ticket to track individually, but if it is easy to fix - let's include it now

src/HotSwap/HotSwap.php
```diff
@@ -0,0 +1,348 @@
...
+
+    /**
+     * Returns the sorted list of implemented interface entry addresses
+     *
+     * @return list<int>
+     */
+    private static function interfaceAddressSet(CData $classEntry): array
+    {
+        $addresses       = [];
+        $shapedEntry     = ReflectionClass::fromCData($classEntry)->getClassEntry();
```
lisachenko: this method body better to extract into overriden implementation of ReflectionClass->getInterfaces() + getInterfaceNames()

src/HotSwap/HotSwap.php
```diff
@@ -0,0 +1,348 @@
...
+        $totalInterfaces = $shapedEntry->num_interfaces;
+        $interfaceList   = $shapedEntry->interfaces;
+        assert($totalInterfaces >= 0);
```
lisachenko: drop this assert

src/HotSwap/HotSwap.php
```diff
@@ -0,0 +1,348 @@
...
+
+        return $addresses;
+    }
+
+    /**
+     * Returns the comparable shape of every property the class declares itself
+     *
+     * @return array<string, array{int, int, int}> Property name => [flags, offset, type mask]
+     */
+    private static function ownPropertySurface(CData $classEntry): array
```
lisachenko: this can be simply ReflectionClass->getProperties() + access reflectionProperty()->getCData(), which will return proper shape

src/Memory/SharedMemory.php
lisachenko: drop this method, it just delegates and do nothing, can be simplified

src/Memory/SharedMemory.php
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * Refuses a mutation on an opcache-shared class entry
+     *
+     * @param string $operation Human-readable operation name for the diagnostic
+     *
+     * @throws SharedMemoryException When the class entry is stored in opcache SHM
+     */
+    public static function assertMutableClassEntry(CData $classEntry, string $operation): void
+    {
+        if (self::isImmutableClassEntry($classEntry)) {
```
lisachenko: instead of asserting via static method, just call where needed direct check: if (!$reflectionClass->isImmutable()) {... throw SharedMemoryException::immutableClassMutation($operation);}

src/Memory/SharedMemory.php
```diff
@@ -0,0 +1,109 @@
...
+     * @return CData Writable zend_function pointer now published in the table
+     */
+    public static function copyOutFunctionEntry(CData $functionEntry, HashTable $table, string $lowerKey): CData
+    {
+        $bucketValue = $table->find($lowerKey);
+        if ($bucketValue === null) {
+            throw SharedMemoryException::functionNotPublished($lowerKey);
+        }
+
+        $writableEntry = Core::trackedNew('zend_function', true);
```
lisachenko: it looks like function creation, which can be inside ReflectionFunction/ReflectionMethod as static constructor, check how this entire method can be optimzed/avoided, then entire class can be removed

src/Reflection/FunctionBodySwap.php
```diff
@@ -0,0 +1,556 @@
...
+     *                                   the table's lifetime is guarded by the body refcount alone
+     *                                   (method donors about to be released via destroy_zend_class)
+     * @param bool  $destroyPrevious     False when the previous body must stay allocated on commit
+     *                                   (opcache shared memory is never freed)
+     * @param int   $publishedShares     Number of table buckets pointing at this entry structure,
+     *                                   see countPublishedShares()
+     *
+     * @internal
+     */
+    public static function swapUserFunctionBody(
```
lisachenko, line 120: same proposal with having ReflectionFunction/ReflecionMethod, like in ClassDelta - accept high-level interfaces, incapsulate low-level details around relevant zvals inside relevant classes

src/Reflection/FunctionLikeTrait.php
```diff
@@ -136,35 +128,35 @@ public function redefine(\Closure $newCode): void
...
+            $newFunction        = $closureEntry->getRawFunction();
+
+            $isSharedMemoryEntry = $this->isImmutable();
+            if ($isSharedMemoryEntry) {
+                if ($this->getCommonPointer()->scope !== null) {
+                    throw SharedMemoryException::immutableMethodTable();
+                }
+                // Copy the entry out of SHM: the per-process function-table bucket is
+                // repointed at a writable container, the SHM original stays untouched
+                $this->pointer = SharedMemory::copyOutFunctionEntry(
```
lisachenko: inline here in the Trait this copyOutFunctionEntry - because we operating here

src/Reflection/ReflectionClass.php
```diff
@@ -437,6 +444,86 @@ public function isUserDefined()
...
+     * The declared shape (see phpstan.dist.neon typeAliases and AGENTS.md) is the
+     * single narrowing point for the class-entry fields the runtime mutation
+     * machinery touches. Keep the raw pointer for FFI hand-offs; this view serves
+     * the field access.
+     *
+     * @return ZendClassEntryShape
+     *
+     * @internal shared with the hot-swap machinery (ClassDelta/HotSwap)
+     */
+    public function getClassEntry(): object
```
lisachenko: let's keep getRawValue() + just keep the return type shape without self::asStructView()

src/Reflection/ReflectionClass.php
```diff
@@ -437,6 +444,86 @@ public function isUserDefined()
...
+        return $classEntry;
+    }
+
+    /**
+     * Widens a CData handle to plain `object` so a shape @var can be declared on it
+     *
+     * FFI\CData is final: a shape alias (stdClass&object{...}) is not a subtype of
+     * the CData native type, so the narrowing must go through the object supertype.
+     */
+    private static function asStructView(CData $struct): object
```
lisachenko: not needed, drop it

src/Reflection/ReflectionClass.php
```diff
@@ -483,6 +570,7 @@ public function getBackedEnumTable(): ?HashTable
      */
     public function removeMethods(string ...$methodNames): void
     {
+        SharedMemory::assertMutableClassEntry($this->pointer, 'remove methods');
```
lisachenko: better to move this check internally and use it dynamic here: if (!$this->isImmutable())

src/Reflection/ReflectionClass.php
```diff
@@ -928,13 +1016,7 @@ private static function parseTraitMethodReference(string $traitMethod): array
      */
     private function assertTraitConfigurationIsMutable(): void
```
lisachenko, line 1224: can be dropped, there is one single isImmutable() method to check

src/Reflection/ReflectionClassConstant.php
```diff
@@ -76,6 +76,40 @@ public static function fromCData(CData $constantEntry, string $constantName): Re
...
+     * donor entries) cannot be wrapped through fromCData(), which resolves the
+     * native reflection state by name.
+     *
+     * @param CData $constantEntry zend_class_constant pointer
+     *
+     * @return ZendClassConstantShape
+     *
+     * @internal shared with the hot-swap machinery (ClassDelta)
+     */
+    public static function viewConstantEntry(CData $constantEntry): object
```
lisachenko: set name getRawValue() +  make it dynamic + keep the object shape return type

src/Reflection/ReflectionClassConstant.php
```diff
@@ -76,6 +76,40 @@ public static function fromCData(CData $constantEntry, string $constantName): Re
...
+        return $shapedEntry;
+    }
+
+    /**
+     * Widens a CData handle to plain `object` so a shape @var can be declared on it
+     *
+     * FFI\CData is final: a shape alias (stdClass&object{...}) is not a subtype of
+     * the CData native type, so the narrowing must go through the object supertype.
+     */
+    private static function asStructView(CData $struct): object
```
lisachenko: not needed, drop it

src/Reflection/ReflectionProperty.php
```diff
@@ -78,6 +78,40 @@ public static function fromCData(CData $propertyEntry): ReflectionProperty
...
+     * hot-swap donor entries) cannot be wrapped through fromCData(), which resolves
+     * the native reflection state by name.
+     *
+     * @param CData $propertyInfo zend_property_info pointer
+     *
+     * @return ZendPropertyInfoShape
+     *
+     * @internal shared with the hot-swap machinery (HotSwap/ClassDelta)
+     */
+    public static function viewPropertyInfo(CData $propertyInfo): object
```
lisachenko: to normal method getRawValue() + object shape in return section

src/Reflection/ReflectionProperty.php
```diff
@@ -78,6 +78,40 @@ public static function fromCData(CData $propertyEntry): ReflectionProperty
...
+        return $shapedInfo;
+    }
+
+    /**
+     * Widens a CData handle to plain `object` so a shape @var can be declared on it
+     *
+     * FFI\CData is final: a shape alias (stdClass&object{...}) is not a subtype of
+     * the CData native type, so the narrowing must go through the object supertype.
+     */
+    private static function asStructView(CData $struct): object
```
lisachenko: drop static method eniterely

src/Reflection/ReflectionValue.php
```diff
@@ -464,6 +464,39 @@ public function getRawValue(): CData
...
+        return $valueEntry;
+    }
+
+    /**
+     * Widens a CData handle to plain `object` so a shape @var can be declared on it
+     *
+     * FFI\CData is final: a shape alias (stdClass&object{...}) is not a subtype of
+     * the CData native type, so the narrowing must go through the object supertype.
+     */
+    private static function asStructView(CData $struct): object
```
lisachenko: drop static method

src/Type/StructArray.php
```diff
@@ -0,0 +1,141 @@
...
+    /**
+     * Raw view over an array of structures or pointers (interface lists, single
+     * struct dereferences): elements stay plain CData handles
+     *
+     * @param CData       $structTable Typed base pointer
+     * @param int<0, max> $count       Number of elements behind the pointer
+     *
+     * @return self<CData>
+     */
+    public static function ofStructs(CData $structTable, int $count): self
```
lisachenko: this not needed, just give it to default constructor + put default for the phpstan template to CData, so it will be used by default if not specified explicitly somewhere

src/Type/StructArray.php
```diff
@@ -0,0 +1,141 @@
...
+        $element = $this->asStructView($this->rawAt($index));
+
+        return $element;
+    }
+
+    /**
+     * Widens a CData handle to plain `object` so the element shape @var can be
+     * declared on it (FFI\CData is final and cannot intersect object shapes)
+     */
+    private function asStructView(CData $element): object
```
lisachenko: drop method

src/Type/StructArray.php
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * Returns the element at the given index, resolved to the element shape
+     *
+     * @return T
+     */
+    public function at(int $index): object
+    {
+        /** @var T $element */
+        $element = $this->asStructView($this->rawAt($index));
```
lisachenko: just operate on template-level and generic, do not need to wrap this into special asStructView - phpstan infers this statically, we dont need this dynamically when method is executed

src/Type/StructArray.php
```diff
@@ -0,0 +1,141 @@
...
+        return $element;
+    }
+
+    /**
+     * Returns the element at the given index as the raw CData handle
+     *
+     * Use this form when the element travels into FFI primitives that require
+     * CData arguments (Core::addr()/memcpy()/cast()).
+     */
+    public function rawAt(int $index): CData
```
lisachenko: this should support template already + object shape return, update phpDoc block

src/Type/StructArray.php
```diff
@@ -0,0 +1,141 @@
...
+    {
+        if ($index < 0 || $index >= $this->count) {
+            throw new \OutOfBoundsException(
+                "Struct array index {$index} is out of bounds, valid range is 0..{$this->count}",
+            );
+        }
+        $element = $this->baseAddress[$index];
+        assert($element instanceof CData);
```
lisachenko: drop this assert

AGENTS.md
```diff
@@ -137,6 +137,28 @@ inside the module; consumers see plain PHP values and framework wrapper objects
...
 audited for it.
 
+## Engine structs are typed by shape, not by call-site asserts
+
+Inside the core layer, engine structs are described ONCE as named
+[PHPStan object shapes](https://phpstan.org/writing-php-code/phpdoc-types#object-shapes)
+via `parameters.typeAliases` in `phpstan.dist.neon` (`ZendFunctionCommonShape`,
+`ZendClassEntryShape`, `ZvalShape`, ...). The shaped accessor lives on the
+reflection/type class that OWNS the struct - never in a standalone helper class:
+`FunctionLikeTrait::getCommonPointer()/getOpArrayPointer()` for zend_function,
```
lisachenko: update docs according to new suggested names

lisachenko:
Full encapsulation pass done in a53be8a..87a036f (3 commits). The whole raw-CData surface now lives behind the reflection/type classes; `ClassDelta`/`HotSwap`/`FunctionBodySwap` operate on `Reflection*` objects.

**High-level APIs added / used:**
- `ReflectionClass` now overrides `getMethods()`/`getReflectionConstants()`/`getReflectionConstant()`/`hasReflectionConstant()`/`getProperties()`/`getInterfaces()`+`getInterfaceNames()` from the low-level tables — emulating `Reflection*` even for dynamic in-memory methods/constants residing as structures.
- Comparison encapsulated as `ReflectionMethod::equals()`, `ReflectionValue::equals()`, `ReflectionClassConstant::equals()`; declaring-class checks use `getDeclaringClass()`.
- Shapes moved onto the existing `getRawValue()`/`getRawData()` (dynamic, with `@return …Shape`); the `asStructView()` wrappers and the extra static view methods are gone — PHPStan infers the shapes statically.
- `StructArray` gained `ArrayAccess` + `replace(int, mixed): mixed`, defaults its `@template T` to `FFI\CData`, and dropped the named-constructor helper/asserts.
- `ClassDelta`/`FunctionBodySwap` accept `ReflectionClass`/`ReflectionFunction`/`ReflectionMethod` instead of CData.

**SharedMemory is deleted** — the copy-out moved to the reflection/trait side, and the guard is now a direct `if (!$reflectionClass->isImmutable()) throw SharedMemoryException::immutableClassMutation(...)` at the call site. `SharedMemoryException` stays in the `Memory` namespace, thrown via its static constructors.

**Fault injection removed** — `$applyFailureInjector` and the two fake-failure tests are gone, replaced by one test that exercises a **genuine** collision rollback. That surfaced a real latent bug I fixed: `HashTable::add()` used `HASH_ADD_NEW`, which in a non-debug build skips the duplicate check and silently inserts a second bucket under the same key (and it compared the result against `Core::FAILURE`/`-1`, but the engine returns `null` on failure). Switched to `HASH_ADD` + `=== null`, matching the existing `addIndex()` pattern — so duplicate detection (and the rollback test) is now real.

**Exceptions**: remaining inline `new` throws converted to static named constructors; AGENTS.md updated with both the struct-ownership rule (shapes on `getRawValue()`, accessors on owning classes, `StructArray` for contiguous arrays) and the exception-static-constructor rule.

**eval() concern (`HotSwap::prepare`)**: split out to #110 (safe AST/compile donor path) per your "separate ticket is fine" note; `prepare()` carries an interim-mechanism docblock referencing it.

Gates: 333 tests OK (net −1: two fault-injection tests → one genuine-rollback test), PHPStan level max clean on a cold cache with zero baseline growth, cs clean, worker-loop soak OK.

---

lisachenko:
The debug-build internal group caught a real memory leak (7 allocations in the hot-swap churn scenario) that the release build silently tolerates. Root-caused under the debug image with valgrind — two distinct ownership bugs, now both fixed (57e756f..7eff1f9):

**Bug 1 — adopted container leaked on a failed publish** (5 of 7 leaks: op_array opcodes, opcode block, compile arena, 2 literal strings). In `ClassDelta::apply()`, the added-method/added-constant loops registered their rollback undo *after* the `addFunctionEntry`/`publishConstantPointer` call — which throws on a duplicate table key. On that throw, the container adopted just before it (holding an op_array body ref, run-time cache, and owned name ref) was never released. Fixed by staging the container-release undo *before* the throwing publish, with the bucket-delete as a separate later-reversed undo entry. This window only became reachable because the round's new genuine-rollback test uses string-literal sources and actually hits the duplicate-key failure (the old fault injector fired before any allocation, so it never exercised it).

**Bug 2 — scanner source buffer leaked for interned parse input** (2 leaks). `zend_prepare_string_for_scanning` extends the source into a fresh rc=1 `zend_string` even for interned literals; `Compiler::parseString()` relied on `acquireReference()` to own it, but that no-ops on a non-refcounted payload, so `release()` never freed the padded copy. Pre-existing (a53be8a leaks it identically in isolation) but only gate-visible now because the rollback deltas parse literal sources (the churn loop uses `sprintf`, non-interned, already fine). Fixed with `ReflectionValue::claimReference()` — records ownership of the reference the engine already placed, without adding one — called after the prepare step.

**Debug build (report_memleaks=1):** churn scenario 7 → 0 leaks; isolated literal `parseString` 1 → 0; full internal group 138 tests, zero leaks. **Release gates:** 333 tests OK, PHPStan level max clean (cold cache, zero baseline growth), cs clean.

The earlier coding-standards failure on this PR is also fixed (57e756f — the refactor reintroduced version-sensitive operator alignment in two `ClassDelta` methods; collapsed to `array_filter()`).

---

AGENTS.md
```diff
@@ -137,6 +137,49 @@ inside the module; consumers see plain PHP values and framework wrapper objects
...
+  analyser infers statically). `FFI\CData` is `final`, so a shape can NOT be attached to
+  a plain `: CData` return and there is no runtime `asStructView()` wrapper - the shape is
+  a return annotation only. If a field is missing from a shape, extend the alias in the
+  config; never spell a shape out inline.
+- **Contiguous struct arrays go through `Type\StructArray`.** zval tables (op_array
+  literals, class default property/static tables), pointer lists (resolved interfaces)
+  and single-struct dereferences use the `ArrayAccess`/`Countable` `StructArray` view
+  (`replace()` for in-place slot overwrite) instead of hand-rolled pointer arithmetic.
+  `newEntry(IS_PTR, ...)` stores the ADDRESS of the CData it is handed, so it needs the
+  dereferenced struct (`StructArray(..., 1)->rawAt(0)`), never the 8-byte pointer.
```
lisachenko: Don’t like this version, better to use ReflectionValue for that instead of using array to take just first element

AGENTS.md
```diff
@@ -137,6 +137,49 @@ inside the module; consumers see plain PHP values and framework wrapper objects
...
 
+## Engine structs are owned by their reflection/type class, never poked from call sites
+
+Every raw engine struct is reached through the ONE reflection/type class that owns it,
+never from a caller reaching into a `CData`. The owning class exposes typed accessors -
+`ReflectionMethod::equals()`, `ReflectionClassConstant::getAccessFlags()`,
+`ReflectionProperty::getOffset()/getFlags()/getSurface()`,
+`ReflectionValue::getBaseType()/equals()/replaceWith()`, `ReflectionClass::getFlags()/
+getParentAddress()/getInterfaceAddresses()/isImmutable()` - and the field pokes
+(`$this->pointer->...`) plus their single-hop narrowing `assert()`s live INSIDE those
```
lisachenko: Remove this statement about “assert”. Not needed when we use phpstan shapes

AGENTS.md
```diff
@@ -137,6 +137,49 @@ inside the module; consumers see plain PHP values and framework wrapper objects
...
+methods. Consumers (`HotSwap`, `ClassDelta`, `FunctionBodySwap`) operate on
```
lisachenko: This applies to all classes - if given class is responsible for that structure, then all external manipulation with this structure should be only via interface API.

src/Type/StructArray.php
```diff
@@ -0,0 +1,130 @@
...
+        private CData $baseAddress,
+        private int $count,
+    ) {}
+
+    public function offsetExists(mixed $offset): bool
+    {
+        return is_int($offset) && $offset >= 0 && $offset < $this->count;
+    }
+
+    public function offsetGet(mixed $offset): CData
```
lisachenko: Can we put int here instead of mixed?

lisachenko: To drop assert

src/Type/StructArray.php
```diff
@@ -0,0 +1,130 @@
...
+    {
+        assert(is_int($offset));
+
+        return $this->rawAt($offset);
+    }
+
+    /**
+     * Overwrites the element at the given index, dropping the previous value
+     */
+    public function offsetSet(mixed $offset, mixed $value): void
```
lisachenko, line 94: Let’s make offset int

src/Type/StructArray.php
lisachenko: Drop method, we have already ArrayAccess API, use it and make entire class template/generic with T, where caller can specify an object shape or default to CData if not specified

src/Type/StructArray.php
```diff
@@ -0,0 +1,130 @@
...
+        return $element;
+    }
+
+    public function count(): int
+    {
+        return max($this->count, 0);
+    }
+
+    /**
+     * @return Traversable<int, CData>
```
lisachenko: Value is not CDdata, it should be T, T can be accepted outside and defaults to CData only in class level block

src/Type/StructArray.php
```diff
@@ -0,0 +1,130 @@
...
+    public function offsetGet(mixed $offset): CData
+    {
+        assert(is_int($offset));
+
+        return $this->rawAt($offset);
+    }
+
+    /**
+     * Overwrites the element at the given index, dropping the previous value
+     */
```
lisachenko, line 93: Method should be templated with T $value

src/Type/StructArray.php
lisachenko: Need to have templated return type to support phpstan generic return type

src/System/ExecutionData.php
```diff
@@ -121,6 +121,28 @@ public function getFunction(): \ReflectionFunctionAbstract
...
+     *
+     * @internal used by FunctionBodySwap to detect entries with live frames
+     */
+    public function getFunctionEntry(): ?ReflectionFunction
+    {
+        $rawFunction = $this->pointer->func;
+        if ($rawFunction === null) {
+            return null;
+        }
+        assert($rawFunction instanceof CData);
```
lisachenko: Drop extra assert, info about type should be from the object shape

src/Reflection/ReflectionValue.php
```diff
@@ -464,6 +469,175 @@ public function getRawValue(): CData
...
+    }
+
+    /**
+     * Returns the base type byte (IS_LONG, IS_STRING, ...) of this zval
+     */
+    public function getBaseType(): int
+    {
+        $this->assertNotReleased();
+        $u1 = $this->pointer->u1;
+        assert($u1 instanceof CData);
```
lisachenko: Drop all asserts - info should be from the object shape in phpstan

src/Reflection/ReflectionValue.php
lisachenko, line 143: Declare the shape here with phpstan docblock to avoid asserts everywhere

src/Reflection/ReflectionValue.php
```diff
@@ -464,6 +469,175 @@ public function getRawValue(): CData
...
+        $type = $typeView->type;
+        assert(is_int($type));
+
+        return $type;
+    }
+
+    /**
+     * Returns the value union (zend_value) of this zval as a raw CData
+     */
+    private function payloadUnion(): CData
```
lisachenko: Drop method - it does nothing important, you can access structure already via shape

src/Reflection/ReflectionProperty.php
```diff
@@ -79,11 +79,98 @@ public static function fromCData(CData $propertyEntry): ReflectionProperty
...
+        $typeMask = $type->type_mask;
+        assert(is_int($typeMask));
+
+        return $typeMask;
+    }
+
+    /**
+     * Returns the numeric address of the class entry that declares this property
+     */
+    public function getDeclaringClassAddress(): int
```
lisachenko: Better to use getDeclaringClass emulation which takes pointer->ce, creates instance of ReflectionClass and then caller will call getAddress on it

src/Reflection/ReflectionClassConstant.php
```diff
@@ -76,6 +76,152 @@ public static function fromCData(CData $constantEntry, string $constantName): Re
...
+        $accessFlags = $extra->constant_flags;
+        assert(is_int($accessFlags));
+
+        return $accessFlags;
+    }
+
+    /**
+     * Returns the numeric address of the declaring class entry of this constant
+     */
+    public function getDeclaringClassAddress(): int
```
lisachenko: Just emulate getDeclaringClass() and getAddress will be called on it then

src/Reflection/ReflectionClassConstant.php
```diff
@@ -76,6 +76,152 @@ public static function fromCData(CData $constantEntry, string $constantName): Re
...
+
+        return $reflectionConstant;
+    }
+
+    /**
+     * Returns the raw zend_class_constant pointer this reflection wraps
+     *
+     * The pointer is a live view into engine memory; prefer the typed accessors
+     * (getReflectionValue(), equals(), getDeclaringClass()) over poking fields.
+     */
```
lisachenko, line 104: Add return object shape for phpstan inside phpDoc block - and to the field itself

src/Reflection/ReflectionClass.php
```diff
@@ -437,6 +445,333 @@ public function isUserDefined()
...
+
+        return $declaredMethods;
+    }
+
+    /**
+     * Returns a single declared/inherited method by name, or null when not present
+     *
+     * @internal shared with the hot-swap machinery (ClassDelta)
+     */
+    public function findMethod(string $name): ?ReflectionMethod
```
lisachenko: Just override the internal getMethod()/hasMethod() reflection API, it will be compatible with native reflection method

src/Reflection/ReflectionClass.php
```diff
@@ -437,6 +445,333 @@ public function isUserDefined()
...
+     *
+     * Reads the low-level method table, so it also sees dynamic/non-persistent methods
+     * that live only as structures in memory; inherited entries (whose scope is another
+     * class) are excluded. Each value is a pointer-level ReflectionMethod.
+     *
+     * @return array<string, ReflectionMethod>
+     *
+     * @internal shared with the hot-swap machinery (ClassDelta)
+     */
+    public function getDeclaredMethods(): array
```
lisachenko, line 542: This can be simply overridden getMethods() API. And consumers should filter if internal method or not

src/Reflection/ReflectionClass.php
```diff
@@ -437,6 +445,333 @@ public function isUserDefined()
...
+    }
+
+    /**
+     * Returns the sorted numeric addresses of the resolved interface entries
+     *
+     * @return list<int>
+     *
+     * @internal used by the hot-swap compatibility guard (linked classes only)
+     */
+    public function getInterfaceAddresses(): array
```
lisachenko: Prefer to mimic/override internal getInterfaces()/getInterfaceNames() implementation. Consumers will iterate and call ReflectionClass instance getAddress() to collect pointers if needed.

src/Reflection/ReflectionClass.php
```diff
@@ -437,6 +445,333 @@ public function isUserDefined()
...
+    {
+        return $this->constantsTable;
+    }
+
+    /**
+     * Returns the numeric address of the parent class entry, or null when unlinked/root
+     *
+     * @internal used by the hot-swap compatibility guard
+     */
+    public function getParentAddress(): ?int
```
lisachenko: Same, use getParent override + getAddress on returned object

src/Reflection/ReflectionClass.php
```diff
@@ -437,6 +445,333 @@ public function isUserDefined()
...
+    }
+
+    /**
+     * Returns the raw zend_class_entry pointer this reflection wraps
+     *
+     * The pointer is a live view into engine memory; prefer the typed accessors on
+     * this class over poking fields on the result.
+     *
+     * @internal shared with the hot-swap machinery (ClassDelta/HotSwap)
+     */
```
lisachenko, line 482: Add to the docblock the phpstan object shape and to property itself

lisachenko:
Round-4 batch addressed in 7eff1f9..d5b0946 (4 commits):

- **StructArray fully generic**: `@template T of object = \FFI\CData` at class level, `ArrayAccess<int,T>`/`IteratorAggregate<int,T>`, `replace(int $offset, T $value): T` templated both ways, `at()`/`rawAt()` helpers dropped (callers use `$structArray[$i]`), all internal asserts removed. One language limit: PHP forbids narrowing `ArrayAccess::offset*(mixed)` to `int` natively (fatal), so the `int` offset lives in the phpdoc.
- **Native reflection overrides + getAddress()**: `getMethod()` override (covariant return, low-level table), new `hasMethod()` override; the compat gu

---

lisachenko:
Round-4 batch applied in 7eff1f9..d5b0946 (4 commits).

- **`StructArray` fully generic**: `@template T of object = \FFI\CData`, `@implements ArrayAccess<int,T>`/`IteratorAggregate<int,T>`. Element reads (`$arr[$i]`, iteration) and `replace()`'s value are typed `T`; the `at()`/`rawAt()` helpers are gone (callers use `[$i]`); all internal asserts removed.
- **Native reflection overrides + `getAddress()`**: `ReflectionClass::getMethod()`/`hasMethod()` overridden from the low-level table (replaces the bespoke `findMethod()`); compat guard uses `getParentClass()?->getAddress()` and `getInterfaces()` addresses; the `getParentAddress()`/`getInterfaceAddresses()`/`getDeclaringClassAddress()` accessors dropped in favor of `getDeclaringClass()->getAddress()`.
- **Asserts dropped via nested zval shapes** (`ZvalShape`/`ZvalU1Shape`/`ZvalTypeShape`/…) in `ReflectionValue`, `ExecutionData`, `ReflectionClassConstant`. Assert count in touched files 104 → 82.
- **AGENTS.md** generalized: the ownership rule now applies to all classes, prefers native-override + `getAddress()`, drops the assert sentence, and the example uses `$structArray[$i]` + owning-reflection wrapping.

Three items I did **not** force, with reasons — flagging for your call:

1. **`getDeclaredMethods/Constants/Properties` → native `getMethods()`/`getReflectionConstants()`/`getProperties()` rename**: deferred. These run on the *unpublished* donor class, whose dynamic in-memory methods have a scope not in the class table. The native `getMethods()` override uses `fromCData()` (native init), which **throws** for donor-only entries; making it work needs `fromRawEntry()`, which strips native introspection and would regress `getMethods()`/`getProperties()` for normal-class callers — a native-contract break. The custom `getDeclared*` helpers still encapsulate the raw access on `ReflectionClass` (core rule honored); only the native-name rename is held back.
2. **Object shape on `getRawValue(): CData` + backing property** (ReflectionClass/ReflectionClassConstant): not applied — `FFI\CData` is `final`, so the `stdClass&object{…}` shapes aren't CData subtypes and annotating a `: CData` return/property yields `return.phpDocType`/`varTag.nativeType` errors, plus the result is consumed as raw CData (memcpy/newEntry). Dropped the dependent asserts via local shape narrowing instead and kept `getRawValue(): CData`.
3. **`StructArray` ArrayAccess `int` offset**: PHP forbids narrowing `ArrayAccess::offset*(mixed)` to `int` (fatal "must be compatible"); used `@param int` phpdoc + dropped the runtime `is_int` assert — the closest possible.

Gates: 333 tests OK; PHPStan level max clean (cold cache, zero baseline growth); cs clean. Debug leak gate re-verified (churn scenario + internal group 138 tests, **0 leaks**).

---

PR #111, opened 2026-08-05
feat(opcache): binary file API — read, patch and refresh opcache file-cache

lisachenko:
## What & why

z-engine could mutate live engine structures (hot-swap, `redefine()`, `ClassDelta`) but had **no way to touch opcache itself** — all opcache interaction was defensive (`ZEND_ACC_IMMUTABLE` detection, copy-out of SHM, `SharedMemoryException` rejections).

This PR adds a first-class API for **opcache file-cache binary files** (`opcache.file_cache` `.bin`) under a new `ZEngine\OpCache` namespace: read them (any build), deserialize the payload onto the existing engine-struct wrappers, mutate through the framework, **re-serialize a valid binary for the current build**, and refresh the cache so the engine loads and executes the patched code. This is the foundation for AOP, transpiling and source-code protection on top of the file cache, and the groundwork for the future SHM-mutation + hot-swap wiring.

The engine's file-cache format is position-independent (interior pointers stored as byte offsets, interned strings as tagged offsets into an appended string section). `PayloadRelocator` is a faithful function-by-function port of `ext/opcache/zend_file_cache.c`'s `unserialize` (→ real pointers, so the existing wrappers walk the image) and `serialize` (→ a fresh valid binary, re-emitting the string section so **size-changing** edits round-trip). Fields that only matter for execution — opline operands/handlers, `ZEND_MAP_PTR` slots, string GC flags — are preserved byte-for-byte; opcache re-derives them on load.

## What's in it

- **`ZEngine\OpCache`** (`src/OpCache/`):
  - `SystemId` — the build fingerprint (`zend_system_id`, newly exported).
  - `CacheMetaInfo` — immutable `zend_file_cache_metainfo` header codec (parsed through the generated FFI struct, so layouts.json guards the offsets; adler32 via `hash()`).
  - `BinaryCacheFile` — `read` / `locate` / `compile` (via opcache's own serializer in a child) / `script()` / `save()` (atomic, checksum recomputed) / `refresh()` (save + `opcache_invalidate`).
  - `PayloadRelocator` — the unserialize/serialize port (`@internal`).
  - `PersistentScriptView` — facade handing out `ReflectionFunction` / `ReflectionClass` / `HashTable` over the relocated script (no CData in public signatures).
  - `OpCacheException` — typed failure modes via static factories.
  - `SharedMemoryException` moved here from `ZEngine\Memory`; new `SharedMemory` facade centralizes immutability detection (and is the intended home for future SHM binding).
- **Generator** (`tools/generator/`): slices `zend_persistent_script`, `zend_early_binding`, `accel_time_t` and `zend_file_cache_metainfo` from `ext/opcache` into `supplement.h`, pulls `zend_script` from the installed Optimizer header, exports `zend_system_id`, and fixes array-typed global emission. `include/8.4/linux-x64-nts/` regenerated (reproduces byte-for-byte; `header-drift` re-verifies).
- **`Core::systemId()`**.
- **Docs**: `docs/opcache-binary.md`, README section, and corrected the stale `ZEngine\Memory\SharedMemory` references.

## Testing

`tests/OpCache/` (child-process pattern, skips when opcache is unavailable):

- Header codec, `SystemId`, `locate`/`compile`/`read`/`verifyChecksum`, foreign-build rejection.
- **Byte-identity**: `relocate()` → `derelocate()` reproduces the payload byte-for-byte across functions, constants, typed properties, attributes (incl. constant-expression args), static vars, try/catch and enums.
- **End-to-end**: patch an integer literal (`41`→`42`) and a **size-changing** string literal, `save()`/`refresh()`, and a fresh worker with `opcache.file_cache_only=1` executes the patched value — proving the engine ran the edited binary.

`composer test` (366 pass), `composer phpstan` (level max, clean), `composer cs:check` (clean). The relocator's unavoidable multi-hop CData reads are scoped to that one audited file in `phpstan.dist.neon`.

## Scope & limits (v1)

- Targets the bundled 64-bit non-Windows build (asserts and fails loudly otherwise).
- Strict, never silent: unsupported structures (intersection/union type lists, property hooks, iterator/ArrayAccess funcs, trait-using classes, compile warnings) raise `unsupportedPayload` rather than writing a subtly corrupt binary.
- **Deferred**: loading patched binaries into shared memory (ZCSG), and applying a patched image to already-loaded classes via `redefine()`/`ClassDelta` — the natural next step that connects this to the hot-swap feature.

---
_Generated by [Claude Code](https://claude.ai/code/session_012aZBexHk1tKyXe2nfWr5mE)_


src/OpCache/BinaryCacheFile.php
```diff
@@ -0,0 +1,249 @@
...
+
+    /**
+     * Reads a cache binary from disk. Works for binaries from any engine
+     * build - only the payload interpretation requires a build match.
+     *
+     * @param string|null $scriptPath The source script this binary caches (used by refresh())
+     */
+    public static function read(string $binPath, ?string $scriptPath = null): self
+    {
+        if (!is_file($binPath)) {
```
lisachenko: better to replace with is_readable() to ensure that file exists and we can access it as well

src/OpCache/BinaryCacheFile.php
```diff
@@ -0,0 +1,249 @@
...
+            throw OpCacheException::binFileNotFound($binPath);
+        }
+        $bytes    = (string) file_get_contents($binPath);
```
lisachenko: instead of casting to string - implement proper handling of errors to ensure that our code is protected from unexpected read errors - what if another process/thread is writing/refreshing this file right now?

lisachenko: on a highload this can be dangerous without advice locking

src/OpCache/BinaryCacheFile.php
```diff
@@ -0,0 +1,249 @@
...
+     * itself would cache, ready for inspection and patching.
+     */
+    public static function compile(string $scriptPath, string $fileCacheDir, ?string $phpBinary = null): self
+    {
+        $realPath = realpath($scriptPath);
+        if ($realPath === false) {
+            throw OpCacheException::scriptNotFound($scriptPath);
+        }
+        // Opcache refuses to boot file_cache_only against a missing directory
+        if (!is_dir($fileCacheDir) && !mkdir($fileCacheDir, 0777, true) && !is_dir($fileCacheDir)) {
```
lisachenko: 0777 - is very wide privilege, we need to separate directory/file creation outside. mask should be configured outside and don't allow all to rwx - this is security issue, use security skill to review and fix this properly

src/OpCache/BinaryCacheFile.php
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    /**
+     * Produces a cache binary for the current build by compiling the script
+     * with opcache's own serializer in a child process, then reads it back.
+     *
+     * This is the "generate" primitive: the result is bit-exact what opcache
+     * itself would cache, ready for inspection and patching.
+     */
+    public static function compile(string $scriptPath, string $fileCacheDir, ?string $phpBinary = null): self
```
lisachenko: what if we will simply add the symfony/console as a dependency, and use it API to run the process

src/OpCache/BinaryCacheFile.php
```diff
@@ -0,0 +1,249 @@
...
+        $realPath = realpath($scriptPath);
+        if ($realPath === false) {
+            throw OpCacheException::scriptNotFound($scriptPath);
+        }
+        // Opcache refuses to boot file_cache_only against a missing directory
+        if (!is_dir($fileCacheDir) && !mkdir($fileCacheDir, 0777, true) && !is_dir($fileCacheDir)) {
+            throw OpCacheException::compilationFailed($realPath, "cannot create the cache directory {$fileCacheDir}");
+        }
+        $command = [
+            $phpBinary ?? PHP_BINARY,
```
lisachenko, line 138: there can be cases when extra options will be needed - please define ability to add more args if needed for console command

src/OpCache/BinaryCacheFile.php
```diff
@@ -0,0 +1,249 @@
...
+            return $this->view;
+        }
+        if (!$this->matchesCurrentBuild()) {
+            throw OpCacheException::systemIdMismatch(SystemId::current(), $this->metaInfo->systemId());
+        }
+        $length = strlen($this->payload);
+        // The buffer is kept alive by the relocator it is handed to; the
+        // relocated image points into it, so it must outlive the view
+        $buffer = Core::new("char[{$length}]", false);
+        Core::memcpy($buffer, $this->payload, $length);
```
lisachenko, line 235: can we reuse already existing string ZVal string payload buffer to avoid copying of bytes? And give just $this->payload to PayloadRelocator ?

src/OpCache/BinaryCacheFile.php
```diff
@@ -0,0 +1,249 @@
...
+        return $this->payload;
+    }
+
+    /**
+     * Materializes the payload into a live image and returns a facade over its
+     * zend_persistent_script. Mutations made through the returned wrappers are
+     * written back by save(). The view is cached, so repeated calls share one
+     * image.
+     */
+    public function script(): PersistentScriptView
```
lisachenko: let's call it getPersistentScriptView()

src/OpCache/BinaryCacheFile.php
```diff
@@ -0,0 +1,249 @@
...
+            $this->metaInfo = $this->metaInfo->withStrSize($length - $this->metaInfo->memSize());
+        }
+        $this->metaInfo = $this->metaInfo
+            ->withChecksum(CacheMetaInfo::checksumOf($this->payload));
+        if ($timestamp !== null) {
+            $this->metaInfo = $this->metaInfo->withTimestamp($timestamp);
+        }
+
+        $directory = dirname($target);
+        if (!is_dir($directory) && !mkdir($directory, 0777, true) && !is_dir($directory)) {
```
lisachenko: same here - directory creation with 0777 - is security risk, end should be configurable/adjustable

src/OpCache/PersistentScriptView.php
```diff
@@ -0,0 +1,103 @@
...
+ * (ReflectionFunction, ReflectionClass, HashTable, ReflectionValue), so a
+ * mutation made through them lands in the image and is written back by
+ * {@see BinaryCacheFile::save()}.
+ *
+ * No public method returns a CData (AGENTS.md): callers see framework wrapper
+ * objects only. Constructed exclusively by BinaryCacheFile::script().
+ *
+ * @internal instances come from BinaryCacheFile::script()
+ */
+final class PersistentScriptView
```
lisachenko: Maybe call it ReflectionOpcacheFile? and mimic a little bit API that we have in Reflection classes?

src/OpCache/PersistentScriptView.php
```diff
@@ -0,0 +1,103 @@
...
+{
+    public function __construct(private readonly CData $script) {}
+
+    /**
+     * The cached script's source path (zend_persistent_script.script.filename)
+     */
+    public function scriptName(): string
```
lisachenko: getFileName() for parity with reflection API

src/OpCache/PersistentScriptView.php
```diff
@@ -0,0 +1,103 @@
...
+    {
+        return StringEntry::fromCData($this->script->script->filename)->getStringValue();
+    }
+
+    /**
+     * The main (file-level) op_array as a ReflectionFunction
+     */
+    public function mainOpArray(): ReflectionFunction
```
lisachenko: getScriptFunction(): ReflectionFunction

src/OpCache/PersistentScriptView.php
```diff
@@ -0,0 +1,103 @@
...
+    {
+        return HashTable::fromCData(FFI::addr($this->script->script->class_table));
+    }
+
+    /**
+     * Every user function compiled into the script
+     *
+     * @return list<ReflectionFunction>
+     */
+    public function functions(): array
```
lisachenko: maybe we can have emulation of getFunctions(): ReflectionFunction[] API? kind of reflection-like API

src/OpCache/PersistentScriptView.php
```diff
@@ -0,0 +1,103 @@
...
+
+        return $functions;
+    }
+
+    /**
+     * Every class compiled into the script
+     *
+     * @return list<ReflectionClass>
+     */
+    public function classes(): array
```
lisachenko: maybe we can also have emulation of getClasses(): ReflectionClass[] API? Reflection-like API

lisachenko: also, we can use keys here to create a map[string]=>ReflectionClass

src/OpCache/PersistentScriptView.php
lisachenko: also, we can use keys here to create a map[string]=>ReflectionFunction

src/OpCache/SharedMemory.php
```diff
@@ -0,0 +1,57 @@
...
+ * ReflectionFunction::redefine()) or rejected with {@see SharedMemoryException}.
+ *
+ * This facade centralizes the detection side. Each check delegates to the
+ * reflection object that owns the underlying engine struct (AGENTS.md: struct
+ * access lives on the owning wrapper), so callers reason about "is this in
+ * shared memory?" without touching engine flags directly. It is also the
+ * intended home for the future SHM-loading path, which will connect a patched
+ * cache binary ({@see BinaryCacheFile}) with the running accelerator.
+ */
+final class SharedMemory
```
lisachenko: drop entire class - delegate calls directly to relevant isImmutable() getters

src/OpCache/BinaryCacheFile.php
lisachenko: I kept z-engine dependency-free (it currently requires only `ext-ffi`), so rather than pull in `symfony/process` I hardened the native `proc_open` path instead: it already uses an array argv (no shell), and I added error handling plus an `$extraDirectives` parameter to append extra `-d` ini options. Happy to switch to Symfony Process if you'd prefer the dependency — just say the word.

---

src/OpCache/BinaryCacheFile.php
lisachenko: `relocate()` rewrites the stored offsets to real addresses **in place**, so the relocator needs writable memory — and PHP strings are immutable/copy-on-write, so `$this->payload` can't be relocated directly. The FFI `char[]` is therefore a required writable copy, not an avoidable one; it's the single working buffer (the image points into it and `save()` re-serializes from it). I added a comment at the allocation explaining this.

---

src/OpCache/BinaryCacheFile.php
lisachenko: Since the view class is now `ReflectionOpcacheFile` (per your other note), I named the accessor `getReflection()` to sit consistently with the `Reflection*` family. Let me know if you'd rather keep `getPersistentScriptView()`.

---

src/OpCache/SharedMemory.php
lisachenko: Done — dropped the `SharedMemory` facade. Immutability detection stays on the `ReflectionClass::isImmutable()` / `ReflectionFunction`/`ReflectionMethod::isImmutable()` getters, and the docs now point at those directly. The `SharedMemoryException` move into `ZEngine\OpCache` is unchanged.

---

PR #128, opened 2026-08-05
feat(reflection): address type substitutions by declaration slot

lisachenko:
Adds `SlotSubstitutionMap` / `TypeSlot` as a companion to `TypeSubstitutionMap`, for the userland-generics satellite package (lisachenko/userland-php-generics#1).

## Why

`TypeSubstitutionMap` can only rewrite a type it can **name**. That leaves out every slot declared as a builtin — `mixed` has no name to key on — and makes it impossible to give two slots that share a placeholder different types. `SlotSubstitutionMap` addresses the declaration itself:

```php
$specialized = (new ClassSpecializer())->specialize(SomeTemplate::class, 'App\Specialized\X', null, new SlotSubstitutionMap([
    [TypeSlot::property('value'), 'int'],
    [TypeSlot::parameter('setNamed', 0), '?App\User'],
    [TypeSlot::returnType('getNamed'), '?App\User'],
]));
```

Both maps may be passed to one call; the slot map wins where they overlap. `ReflectionClass::specialize()` grows the same optional parameter. Backwards compatible.

Nullability is taken from the replacement (`?int`) rather than preserved from the slot, because `mixed` implies null and preserving it would quietly make every rewritten slot nullable. A slot-addressed method always duplicates its `arg_info` block, since the block it would otherwise share with the template is the one being written into.

## The finding worth reviewing

Rewriting a type and having it **enforced** turn out to be different things:

| Slot | Class-like declared type | Builtin declared type (`mixed`, `int`, …) |
|---|---|---|
| Property | rewritten and enforced | rewritten and enforced |
| Parameter | rewritten and enforced | **rejected** |
| Return type | rewritten and enforced | **rejected** |

A property write always consults `zend_property_info`, so properties are free whatever they were declared as. For a builtin parameter or return type the compiler resolves the check at compile time and selects a specialized `ZEND_RECV` / `ZEND_VERIFY_RETURN_TYPE` handler — and those opcodes are **shared with the template** by the copy model. So the rewrite shows up in reflection and changes nothing at run time.

I measured this rather than assumed it: `mixed`→`int`, `string`→`int` and `mixed`→`ClassName` parameters all reflect the new type and accept anything at run time, while `A`→`B` parameters and `A`→`int` returns throw `TypeError` correctly.

Handing back a class that looks specialized and silently stops checking is the same failure mode that makes compile-time AST rewriting unsafe, so those two cases are **rejected** with an exception that names the reason and points at declaring the slot with a placeholder type. This is also why the name-keyed path needs no such restriction: a placeholder is by definition class-like, so every slot it can match is already on the generic path.

## Validation

Runs on the source through native reflection rather than pointer arithmetic — the questions being asked (does this member exist, is it declared here, is its type a single type, does its default still fit) are exactly the ones reflection answers, and keeping them out of `CData` keeps them trustworthy. All rejections happen before any engine state is modified.

Declared defaults are checked because the engine verifies them at compile time and never again: substituting `mixed $x = null` to `int` would otherwise produce an object whose property violates its own declaration.

## Verification

`composer test`: **388 tests / 3798 assertions OK** (+22 vs master), including engine-enforced property, parameter, return, nullable, variadic, class-type and arg_info-isolation cases plus every rejection path.
`composer cs:check`: clean.
`composer phpstan`: clean for every file this PR touches.

> One note on PHPStan: in my sandbox `composer` resolves phpstan 2.1.39, and on an **unmodified** `master` that already reports 19 errors — the committed `phpstan-baseline.neon` predates it. So the baseline may need regenerating independently of this PR; I deliberately did not touch it.

No generator change and no header regeneration: `_ZEND_TYPE_MASK`, `_ZEND_TYPE_NULLABLE_BIT` and `_ZEND_TYPE_NAME_BIT` all landed with #104.

Docs: `docs/class-specialization.md` gains a "Slot-addressed substitution" section with the enforcement matrix, the rejection table and the support-matrix rows.

---
_Generated by [Claude Code](https://claude.ai/code/session_013QjceGerKddAYLJkkQufQR)_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@
 /.phpunit.result.cache
 /.php-cs-fixer.cache
 /build/
+/composer.sandbox.json
+/composer.sandbox.lock
+/var/
--- a/docs/class-specialization.md
+++ b/docs/class-specialization.md
@@ -46,6 +46,72 @@ enforcement follows the substituted type on the copy only: assigning a mismatche
 to a substituted typed property throws `TypeError` on the specialized class while the
 template keeps its original declaration.
 
+## Slot-addressed substitution
+
+`TypeSubstitutionMap` can only rewrite a type it can *name*. A slot declared `mixed` has no
+name to key on, and two slots that share a placeholder cannot be given different types.
+`SlotSubstitutionMap` addresses the declaration itself instead:
+
+```php
+use ZEngine\Reflection\{ClassSpecializer, SlotSubstitutionMap, TypeSlot};
+
+$specialized = (new ClassSpecializer())->specialize(SomeTemplate::class, 'App\Specialized\X', null, new SlotSubstitutionMap([
+    [TypeSlot::property('value'), 'int'],
+    [TypeSlot::parameter('setNamed', 0), '?App\User'],
+    [TypeSlot::returnType('getNamed'), '?App\User'],
+]));
+```
+
+Both maps may be supplied to one call; where they overlap, the slot map wins. Replacement
+types are written the way PHP writes them (`int`, `?int`, `App\User`, `?App\User`), and -
+unlike the name-keyed path, which preserves whatever nullability the template declared - the
+nullability written here is the nullability the copy gets. That difference is deliberate:
+`mixed` implies null, so preserving it would silently make every rewritten slot nullable.
+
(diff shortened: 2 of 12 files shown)


lisachenko:
Need to rebase to resolve conflicts

PR #135, opened 2026-08-06
fix: make Core::pointerAtAddress allocation-free (leaked ~116 B per call)

lisachenko:
Fixes a process-lifetime memory leak in `Core::pointerAtAddress()`: it minted an owned `T*[1]` FFI slot on every call and returned `$slot[0]` — a view that pins the owning slot forever, so ~116 B leaked per call for **every** caller. Introduced in `a56d59d` when `HashTable::uninitializedBucketData()` switched to this primitive, which made every hashtable construction leak.

**Fix**: `FFI::cast()` reinterprets a plain integer as a pointer value with no backing buffer, so the method collapses to a single allocation-free cast. The `assert($pointer instanceof CData)` is gone because the guarantee is now static (`cast()` returns `CData`; PHPStan level max rejects the redundant runtime check).

**Regression tests** (`tests/CoreMemoryTest.php`): an `addressOf()` round-trip, and a 50k-call loop asserting RSS growth stays under 1 MB — verified to fail (+5.4 MB) with the fix reverted.

**Measured, PHP 8.4.19, 50 000 calls**: request memory +5 832 192 B → **+0 B**; RSS +5 556 kB → **+0 kB**.

**Verification on this branch**: full phpunit 396/396, `--group opcache --fail-on-skipped` 26/26, `--group internal --process-isolation` 138/138, PHPStan clean, cs clean.

**Downstream proof** (lisachenko/php-shared-data-extension#13, which surfaced this): with this fix overlaid on tag 8.4.0, its failing "Reclaiming soak" goes from **+9 211 080 B FAIL** to **+32 048 B PASS** (budget 65 536), resident 3.73 kB/cycle (budget 6 kB).

**Release note**: master (8.5) has the same bug via the cascade lineage — the `8.4 → master` merge-up will carry this fix. A `8.4.1` tag after merge lets the `~8.4.0`-pinned dependents pick it up.

Note: branch `8.4` was auto-deleted when cascade PR #133 merged ("automatically delete head branches" acted on the PR head); it has been restored at its previous SHA. Consider excluding version branches from auto-delete, or the next cascade merge will delete it again.

https://claude.ai/code/session_01HRDZ2XsoVuB5uG4qXKL3ny

---
_Generated by [Claude Code](https://claude.ai/code/session_01HRDZ2XsoVuB5uG4qXKL3ny)_

### DIFF
--- a/src/Core.php
+++ b/src/Core.php
@@ -488,23 +488,27 @@ class Core
     /**
      * Materializes a typed pointer from a numeric address (the inverse of addressOf())
      *
-     * The address is written through an integer view of a fresh pointer slot and the
-     * typed pointer value is read back. This is the explicit form of C pointer
-     * arithmetic (base address plus a signed byte offset), which FFI otherwise only
-     * offers through CData operator overloads.
+     * FFI::cast() reinterprets a plain integer as a pointer value, which yields the
+     * typed pointer without allocating anything at all - the result is an unowned view
+     * that carries the address and nothing else.
+     *
+     * It must stay allocation-free: this is the primitive behind every "walk to that
+     * engine address" in the framework (hashtable construction alone calls it once per
+     * table), so an owned buffer per call would be a leak with no upper bound. The
+     * previous implementation wrote the address through an integer view of a fresh
+     * `{$type}[1]` slot and returned `$slot[0]`; because that element view pins its
+     * owning slot for as long as it lives, every call permanently retained the slot
+     * (~116 bytes) - the same FFI ownership trap documented on cast() above.
+     *
+     * The result needs no instanceof narrowing anymore: reading `$slot[0]` off a CData
+     * array yielded an untyped value, while cast() is declared to return CData.
      *
      * @param string $type    Pointer type of the result (eg "zend_arg_info *")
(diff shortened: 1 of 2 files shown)


lisachenko:
Please make all branches 8.x not deletable after merge.

lisachenko:
I can't change repository rules with the tooling available in this session (no ruleset/branch-protection API access), so this one needs a minute of your admin UI — two equivalent options:

**Option A — ruleset (recommended):** Settings → Rules → Rulesets → **New branch ruleset**
- Name: `protect-version-branches`, Enforcement: **Active**
- Target branches → Add target → **Include by pattern** → `8.*` (add `0.x` too if you want it kept)
- Under Rules, check **Restrict deletions** only (leave the rest off so cascade merges and fix pushes keep working)

**Option B — classic protection rule:** Settings → Branches → **Add branch protection rule** → pattern `8.*`, save with no checkboxes enabled — matching branches become non-deletable and non-force-pushable by default.

Either one makes "Automatically delete head branches" skip `8.x` branches when a cascade PR (like #133) merges, while feature branches still get cleaned up. `8.4` is already restored at its pre-deletion SHA, so once the rule is in place this can't recur.

---

PR #142, opened 2026-08-07
feat: support PHP 8.4 and 8.5 in parallel from one codebase

lisachenko:
## Summary

One z-engine install now runs on **PHP 8.4 and 8.5 in parallel**. The branch merges `master` (which carried the 8.5 support) into the `8.4` line and unifies the two: both generated definition sets are bundled (`include/8.4/`, `include/8.5/`), `Core::init()` selects the set of the *running* interpreter, and every hand-declared value that moved between minors became a `PHP_VERSION_ID` conditional. z-engine owns all version-dependent header complexity — consumers (e.g. `userland-php-generics`) never see it.

## What changed

**Version-dependent code unified (`src/`)**
- Boot guard widened to `[80400, 80600)`; the refusal message now lists every supported minor.
- `Core::ZEND_ACC_USE_GUARDS` is `0x800` on 8.4 / `0x40000000` on 8.5, expressed as a conditional constant.
- `NodeKind`: 8.5 inserted `AST_OP_ARRAY`/`AST_CAST_VOID` (shifting later kinds), added `AST_PIPE` and removed `AST_CLONE`/`AST_EXIT`. Every shifted kind is a conditional; kinds absent on the running minor get a negative sentinel, and the `name()` reverse map indexes only real kinds.
- `PayloadRelocator`: matches `ZEND_AST_OP_ARRAY` only on 8.5 (kind 66 is `AST_ZNODE` on 8.4), and walks `zend_attribute.validation_error` + the `DECLARE_ATTRIBUTED_CONST` opline pair only where the engine has them.
- `FunctionBodySwap`/`PendingBodySwap`/`ClassDelta`: the 8.4 statics-defaults duplication for closure donors is restored and gated — on 8.4 a closure donor owns (and destroys) its defaults table so the entry must duplicate; since 8.5 the table is guarded by the body refcount alone and duplicating would leak. The ownership rules are documented on `unshareStaticVariables()`.

**Tests**
- The 8.5-only file-cache shapes (attributed constant, const closure) use 8.5-only *syntax*, so they moved out of the shared `answer.php` fixture into `answer-php85.php` with a new `Php85SerializerShapesTest` in group `opcache-php85`, asserted skip-free via `composer test:opcache85` on 8.5. The `opcache` group stays skip-free on every minor.

**Tooling / CI / docs**
- `composer.json`: `php: ~8.4.0 || ~8.5.0`.
- `tools/generator/generate.php` targets **both** minors by default, so `composer gen-headers` and the `header-drift` job cover the whole `include/` tree; `symbols.php` regains the `PHP_VERSION_ID` gate around `zend_ast_op_array` (absent from 8.4 headers).
- CI runs tests, PHPStan and the debug-build internal/opcache groups on a `{8.4, 8.5}` matrix.
- README support matrix, AGENTS.md (version matching + branch model), CONTRIBUTING and the bug-report template describe the parallel model. `8.4` and `master` now carry the same two-minor support; `master` remains where the next minor lands first.

## Verification (all local, in containers matching CI)

| Check | PHP 8.4.24 | PHP 8.5.9 |
|---|---|---|
| Default suite (release build) | ✅ 408 tests | ✅ 409 tests |
| `--group internal` (debug build, isolated) | ✅ 142 tests | — |
| `--group opcache --fail-on-skipped` | ✅ 26 tests | ✅ 26 tests |
| `--group opcache-php85 --fail-on-skipped` | n/a (version-gated) | ✅ 2 tests |
| Worker-loop soak | ✅ +32 B | ✅ +32 B |
| `composer gen-headers` drift | ✅ byte-identical | ✅ byte-identical |
| PHPStan level max / cs:check | ✅ | ✅ |

One pre-existing issue surfaced while testing (not introduced here, reproduces on pristine `8.4`): running the **full default suite** on a *debug* 8.4.24 build trips a `zend_hash` consistency assertion in `ClassDeltaTest::testRemovedOverrideRestoresInheritedMethod`. CI never runs that combination (the default suite runs on release builds; debug jobs run only the internal/opcache groups, which are green) — left out of scope.

## Companion PR

`lisachenko/userland-php-generics` has a matching PR that widens to `php: ^8.4` and `lisachenko/z-engine: 8.4.x-dev || 8.5.x-dev`. Its 8.5 CI legs need this PR merged (and the Packagist mirror refreshed) before they can resolve.

---
_Generated by [Claude Code](https://claude.ai/code/session_0184AgTjKVX4S6qmvRsdM6dt)_


lisachenko:
Please stop here, only relevant consumer library should support several versions in parallel, not the z-engine itself. Technically, master already supports parallel binaries, so not merge back here php8.5 changes

lisachenko:
Understood — closing this, z-engine keeps its one-minor-per-branch model.

The consumer-side goal is already achieved with no z-engine change needed: [userland-php-generics#13](https://github.com/lisachenko/userland-php-generics/pull/13) takes `lisachenko/z-engine: "8.4.x-dev || 8.5.x-dev"`, so composer resolves the `8.4` branch on a PHP 8.4 platform and `master` (through its `8.5.x-dev` branch alias) on PHP 8.5. Its full CI matrix — tests, opcache, debug internal group, PHPStan on both 8.4 and 8.5, plus an experimental 8.6-nightly analysis job — is green with that resolution.

---

PR #164, opened 2026-08-08
feat(core): add ExecutionData::hasThis()/getThis()

### LINKED ISSUE #161: ExecutionData::hasThis() / getThisObject() — encapsulate This.type_info call-info semantics
## Motivation (from zdebug)

`ExecutionData::getThis()` returns the raw `This` zval, whose `u1.type_info` doubles as the **frame's call-info flag word** — it is never a bare `IS_OBJECT` (8). For a method frame it reads as `IS_OBJECT_EX` plus call-info bits (≥ 776). This is a real trap: zdebug shipped a bug from exactly this —

```php
// zdebug src/Context/ContextProvider.php:75 (BUG)
if ($execution->getThis()->getType() === ReflectionValue::IS_OBJECT) { ... } // never true
```

— which silently dropped `$this` from the debugger's `context_get` and made `$this`-based breakpoint conditions never fire.

z-engine already hides this trap for the symbol table (`getSymbolTable()` checks `ZEND_CALL_HAS_SYMBOL_TABLE` against `This.u1.type_info` internally, `src/System/ExecutionData.php:298-304`). The same courtesy is missing for `$this` itself.

## Proposed API

On `System\ExecutionData` (`src/System/ExecutionData.php`):

```php
/**
 * Whether this frame was called with a bound $this
 * (ZEND_CALL_HAS_THIS in the frame's call-info flags).
 */
public function hasThis(): bool;

/**
 * The frame's bound $this as a value entry, or null when the frame
 * has none (function call, static call, main scope).
 */
public function getThisObject(): ?ReflectionValue;
```

## Implementation sketch

- `hasThis()`: reuse the existing private `getThisZvalShape()` helper — `($shape->u1->type_info & Core::engineConstant('ZEND_CALL_HAS_THIS')) !== 0`, following the exact pattern of `getSymbolTable()`. (`ZEND_CALL_HAS_THIS` needs exporting the same way `ZEND_CALL_HAS_SYMBOL_TABLE` already is.)
- `getThisObject()`: `return $this->hasThis() ? ReflectionValue::fromValueEntry(Core::addr($this->pointer->This)) : null;`
- Docblock on `getThis()` gains a warning pointing to the new methods for object access.
- Unit tests: instance method frame (`hasThis() === true`, object identity via native value), static method and plain function frames (`false`/`null`).

lisachenko:
Closes #161

## What

Adds two safe accessors for a frame's bound object scope to `System\ExecutionData`, mirroring the flag-check pattern of `getSymbolTable()`:

- `hasThis(): bool` — tests `ZEND_CALL_HAS_THIS` against the frame's call_info (the `type_info` of the `This` zval, read via the existing `getThisZvalShape()` view and `Core::engineConstant()`).
- `getThis(): ?ReflectionValue` — the bound `$this` as a `ReflectionValue` when the flag is set, `null` for frames without a bound object (plain function, static call, main scope).

Also extends `getThis()`'s docblock with a warning that the `This` zval doubles as the frame's call-info word: its `type_info` is never a bare `IS_OBJECT` (a method frame reads `IS_OBJECT_EX` + call-info bits, 776 and up), and points readers to the new accessors.

## Why

Comparing `getThis()->getType() === IS_OBJECT` to detect a bound `$this` never matches — exactly the trap the zdebug debugger shipped as a real bug. The engine constant `ZEND_CALL_HAS_THIS` (776 on linux-x64) is already exported in `include/8.4/*/constants.php`, so no header regeneration is needed; `include/` is untouched. `src/Reflection/ReflectionValue.php` is deliberately untouched (owned by a sibling task).

## Tests

New unit tests in `tests/System/ExecutionDataTest.php`:

- instance method frame → `hasThis()` is true and `getThisObject()`'s native value is the bound object (`$this` of the test),
- static method frame → false / null,
- plain function frame (static closure) → false / null.

## Testing note

phpstan level max + php-cs-fixer run locally on PHP 8.5.9; the unit test suite was NOT run locally per AGENTS.md version-matching rule (container PHP ≠ branch target 8.4) — CI must run it.

https://claude.ai/code/session_01V5EyK92Zsx1nXuns6wbho9

---
_Generated by [Claude Code](https://claude.ai/code/session_01V5EyK92Zsx1nXuns6wbho9)_

### DIFF
--- a/src/System/ExecutionData.php
+++ b/src/System/ExecutionData.php
@@ -147,10 +147,49 @@ class ExecutionData
      * Returns the current object scope
      *
      * This contains following: this + call_info + num_args
+     *
+     * WARNING: the This zval doubles as the frame's call-info word - its
+     * u1.type_info carries the ZEND_CALL_* frame flags packed next to the type,
+     * so it is never a bare IS_OBJECT: a method frame reads as IS_OBJECT_EX plus
+     * call-info bits (776 and up), and a frame without a bound object still has
+     * flags there. Never compare getThis()->getType() against IS_OBJECT to detect
+     * a bound $this - use hasThis() / getThisObject() instead.
      */
     public function getThis(): ReflectionValue
     {
         return ReflectionValue::fromValueEntry(Core::addr($this->pointer->This));
+    }
+
+    /**
+     * Checks whether this frame has a bound $this object
+     *
+     * The presence of an object scope is flagged by ZEND_CALL_HAS_THIS in the
+     * frame's call_info, which lives in the type_info of the This zval (see the
+     * getThis() warning). Plain function frames, static calls and the main scope
+     * carry no bound object.
+     */
+    public function hasThis(): bool
+    {
+        // ZEND_CALL_INFO(call): the frame flags live in the type_info of This
+        $callInfo = $this->getThisZvalShape()->u1->type_info;
+
+        return ($callInfo & Core::engineConstant('ZEND_CALL_HAS_THIS')) !== 0;
+    }
+
+    /**
+     * Returns the bound $this object of this frame, or null when the frame has none
+     * (plain function, static call, main scope)
+     *
+     * Unlike getThis() this checks the ZEND_CALL_HAS_THIS frame flag first, so it
+     * is the safe way to observe the object scope of an arbitrary frame.
+     */
+    public function getThisObject(): ?ReflectionValue
+    {
+        if (!$this->hasThis()) {
+            return null;
+        }
+
+        return $this->getThis();
     }
 
     /**
--- a/src/System/ExecutionData.php
+++ b/src/System/ExecutionData.php
@@ -144,19 +144,21 @@ class ExecutionData
     }
 
     /**
-     * Returns the current object scope
+     * Returns the bound $this object of this frame, or null when the frame has none
+     * (plain function, static call, main scope)
      *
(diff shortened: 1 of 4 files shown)


src/System/ExecutionData.php
```diff
@@ -147,12 +147,51 @@ public function getFunctionEntry(): ?ReflectionFunction
...
+    }
+
+    /**
+     * Returns the bound $this object of this frame, or null when the frame has none
+     * (plain function, static call, main scope)
+     *
+     * Unlike getThis() this checks the ZEND_CALL_HAS_THIS frame flag first, so it
+     * is the safe way to observe the object scope of an arbitrary frame.
+     */
+    public function getThisObject(): ?ReflectionValue
```
lisachenko: Just bake this into getThis() and change, I don’t want two similar methods

lisachenko: Done in d604139: `getThis(): ?ReflectionValue` now checks `ZEND_CALL_HAS_THIS` itself and returns `null` for frames without a bound `$this`; `getThisObject()` is gone. Kept `hasThis()` as the cheap flag-only predicate (no `ReflectionValue` construction) — happy to drop it too if you'd rather have just the nullable accessor. The one caller reading `getThis()` unconditionally (`UnsetPropertyHook`) now guards the null case explicitly.

---

Issue #119, opened 2026-08-05
feat(opcache): macOS/arm64 support for the relocator (absolute-address oplines)


lisachenko:
Follow-up to the OpCache binary-file API (#111, merged). Relates to #58 (macOS).

The relocator preserves oplines verbatim, relying on `ZEND_USE_ABS_CONST_ADDR == 0` and `ZEND_USE_ABS_JMP_ADDR == 0` (true on linux-x64: operands are literal indexes, jumps are relative). On builds where either is 1, `zend_file_cache_(un)serialize_op_array` relocates `opline->op1/op2.zv` and `jmp_addr` per-opline — the relocator must do the same there.

**Scope:** add the per-opline absolute-address relocation branches for macOS/arm64, gated on the platform, once #58 provides the base platform support. Track alongside that issue.

**Windows is intentionally out of scope** (rescoped 2026-08-19): full opcache support on Windows is a deliberate non-goal for z-engine. The relocator keeps rejecting Windows loudly (`PayloadRelocator` constructor throws `OpCacheException::unsupportedPayload`), and the docs record it as an intentional limitation rather than pending work. The original Windows half of this ticket (per-opline absolute-address branches for the Windows build plus the Windows `zend_file_cache_get_bin_file_path` layout) is preserved here for reference in case that policy ever changes, but no Windows relocator work is planned.

PR #183, opened 2026-08-09
feat: boot from Composer's autoloader (closes #21), plus ClassSpecializer::evict() and Core::sizeOfType()

### LINKED ISSUE #21: Add automatic self-registration via Composer
Autoload can be used for automatic Core initialization, thus simplify usage of this library. Unfortunately, Composer doesn't support automatic preload right now, so only CLI mode can be used for auto-registration right now.

lisachenko:
## Why

An audit of `userland-php-generics` (the reference consumer) found it reaching behind z-engine's API line for operations that had no named entry point: probing `isset(Core::$executor)` to ask "is the engine booted", deleting class-table buckets through the `@internal` `HashTable::delete()`, pairing `Core::sizeof(Core::type(...))` so a raw `FFI\CType` crossed the boundary, and re-deriving "can the engine run here" from `ini_get('ffi.enable')`.

Review feedback then pointed at the better fix for most of that: **don't make consumers initialize the engine at all.**

## Closes #21 — automatic initialization via Composer

`bootstrap.php`, registered through `autoload.files`, boots the bridge. `require vendor/autoload.php` is now the whole boot for every consumer.

Issue #21 has been open since 2019 on one blocker, recorded in its own comments: *"Preloading breaks all the things — because composer autoloader calls the `Core::init()` before preload initialization."* That is real. An unconditional boot binds the definitions with `FFI::cdef()`, which lasts for the preload request only, **and** leaves an engine behind that turns the script's own `Core::preload()` into a no-op — so the server starts and every request after it fails.

So the bootstrap recognises the preload stage and serves it with `Core::preload()`, whose `FFI::load()` publishes the definitions under `FFI_SCOPE` for the life of the server; everything else gets `Core::init()`. The stage is identified by the one fact that distinguishes it: **during preloading, the script named by `opcache.preload` is the first file the process included.**

Verified end to end on a matching PHP line, not just asserted:

```
preload via autoloader  → FFI::scope('ZEngine') resolves in the next request   ✅
preload skipping the boot → "Failed loading scope 'ZEngine'"                    ✅ (control)
```

`Core::preload()` is now idempotent too, so the explicit call an existing `opcache.preload` script makes right after requiring the autoloader stays harmless.

### Failure is silent, and explained where it matters

A host without ext-ffi, with `ffi.enable=0`, on an unsupported PHP minor or platform can still legitimately autoload this package — static analysis, FFI-disabled test jobs, composer-time tooling. Throwing from an autoloaded file breaks all of them at `require`, so the boot is left undone in silence. Nothing is lost: `Core::init()` is idempotent, so code that needs the engine calls it and gets the same explanatory `RuntimeException` the bootstrap swallowed. `ZENGINE_AUTOBOOT=0` skips the boot entirely.

Note z-engine's **own** test bootstrap deliberately still calls `Core::init()` explicitly — every test here drives the engine, so an unsupported host has to say so once rather than through hundreds of confusing failures. `AutoBootTest` proves the automatic path in child processes, where it can be observed without that line interfering.

## Also in this PR

- **`ClassSpecializer::evict(string $className): bool`** — the destroying counterpart of `specialize()`: removes the class-table bucket so `destroy_zend_class()` dismantles the entry *now*, while op_array-refcounted bodies shared with the source stay alive. Reads the class through `ReflectionClass::isUserDefined()/isImmutable()/isPreloaded()` — no `ce_flags` from the outside. `ReflectionClass::isPreloaded()` is added beside `isImmutable()`.
- **`Core::sizeOfType(string $type): int`** — the named form of the `sizeof(type(...))` pair; `Core::type()` is now `@internal`, so no raw `CType` crosses the API line.
- **`AGENTS.md`** — the `@internal` / FFI-CData rule for dependent packages folded into *Public APIs never leak CData*.

Deliberately **no** `ReflectionClass::evict()` mirror: an instance method destroying the entry its own wrapper points at invites use-after-free on `$this->pointer`.

## Review comments

All four addressed in b1c0b92 — `Core::isUsable()` dropped (the bootstrap removed the need for it rather than relocating it), `evict()` moved onto the reflection API, AGENTS.md section folded. Replies are on the threads.

## Notes for review

- Consumer counterpart: lisachenko/userland-php-generics#14 (same branch name), which now boots nothing at all.
- Per the branch model this should merge up to `master` so the 8.5 line carries the same API and bootstrap.
- Verified on this container (PHP 8.5): `phpstan` and `php-cs-fixer` green; auto-boot, opt-out, silent-failure and the preload→request path each exercised directly. The suite itself needs a PHP 8.4 host for this line — covered by CI.
- c8c8dfa skips the two preload cases on Windows, where PHP has no preloading at all (issue #119).

https://claude.ai/code/session_019TGQqR7ByjHrVSYKVHPrkJ

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -211,6 +211,26 @@ inside the module; consumers see plain PHP values and framework wrapper objects
 it or convert it. This is what keeps the FFI blast radius confined to code that is
 audited for it.
 
+## Consuming z-engine from another package
+
+Dependant packages (userland-php-generics is the reference consumer) talk to z-engine through
+its documented public API and nothing else:
+
+- **Lifecycle**: `Core::init()`, `Core::preload()`, `Core::isInitialized()` (never probe
+  `isset(Core::$executor)` — the wrappers are assigned mid-boot, before the layout checks
+  pass, so the probe reports a half-booted bridge as ready), and `Core::isUsable()` for
+  "should the engine paths run in this environment at all" (never re-derive that from
+  `ini_get('ffi.enable')` — a boolean filter rejects the supported `preload` mode).
+- **Services and wrappers**: `ClassSpecializer` (including `evict()`), `HotSwap`,
+  `PersistentHeap`, the `Reflection\*` and `Type\*` wrapper objects, and the substitution
+  value objects (`TypeSubstitutionMap`, `SlotSubstitutionMap`, `TypeSlot`).
+
+Off-limits to dependants, without exception: the engine-global wrappers
+`Core::$executor` / `Core::$compiler` / `Core::$modules`, every method marked `@internal`,
+and anything returning a raw `FFI\CData`/`FFI\CType`. If a dependant needs an operation that
+only exists behind that line — as happened with class-table eviction and struct sizes — the
+fix is a named public API here, not a reach-through there.
+
 ## Engine structs are owned by their reflection/type class, never poked from call sites
 
 This applies to EVERY class: if a class is responsible for a structure, then all external
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -211,25 +211,14 @@ inside the module; consumers see plain PHP values and framework wrapper objects
 it or convert it. This is what keeps the FFI blast radius confined to code that is
 audited for it.
 
-## Consuming z-engine from another package
-
-Dependant packages (userland-php-generics is the reference consumer) talk to z-engine through
-its documented public API and nothing else:
-
-- **Lifecycle**: `Core::init()`, `Core::preload()`, `Core::isInitialized()` (never probe
-  `isset(Core::$executor)` — the wrappers are assigned mid-boot, before the layout checks
-  pass, so the probe reports a half-booted bridge as ready), and `Core::isUsable()` for
-  "should the engine paths run in this environment at all" (never re-derive that from
-  `ini_get('ffi.enable')` — a boolean filter rejects the supported `preload` mode).
-- **Services and wrappers**: `ClassSpecializer` (including `evict()`), `HotSwap`,
-  `PersistentHeap`, the `Reflection\*` and `Type\*` wrapper objects, and the substitution
-  value objects (`TypeSubstitutionMap`, `SlotSubstitutionMap`, `TypeSlot`).
-
-Off-limits to dependants, without exception: the engine-global wrappers
-`Core::$executor` / `Core::$compiler` / `Core::$modules`, every method marked `@internal`,
-and anything returning a raw `FFI\CData`/`FFI\CType`. If a dependant needs an operation that
-only exists behind that line — as happened with class-table eviction and struct sizes — the
-fix is a named public API here, not a reach-through there.
+The same line holds for packages built **on** z-engine. What is off-limits to them is
+every method marked `@internal` and anything handing out a raw `FFI\CData`/`FFI\CType`
+(`Core::type()`, the `getRaw*()` escape hatches) — plus the engine-global wrappers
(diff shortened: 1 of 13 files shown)


AGENTS.md
```diff
@@ -211,6 +211,26 @@ inside the module; consumers see plain PHP values and framework wrapper objects
 it or convert it. This is what keeps the FFI blast radius confined to code that is
 audited for it.
 
+## Consuming z-engine from another package
```
lisachenko: Drop this section, not a primary target for documentation inside z-engine, but I like part about `@internal`, FFI/CData, that can be preserved and merged into next section

src/Core.php
```diff
@@ -366,6 +366,29 @@ public static function isInitialized(): bool
...
+     * Core::preload()), a supported PHP minor, and generated engine definitions for this
+     * platform.
+     *
+     * The non-throwing projection of the boot guard: init() explains a refusal, this one
+     * reports it. Dependants and test bootstraps that need "should the engine paths run
+     * here?" ask this instead of re-deriving the rule from ini_get('ffi.enable') - which is
+     * spelled several different ways and means different things per SAPI, so a hand-rolled
+     * check is wrong somewhere (a boolean filter rejects the supported `preload` mode).
+     */
+    public static function isUsable(): bool
```
lisachenko: Drop this method, Core::init() and Core::isInitialized() should be enough. Don’t want to extend surface too much

lisachenko: All the checks/logic will be inside init() method…

src/Reflection/ClassSpecializer.php
```diff
@@ -128,6 +128,58 @@ class_exists($sourceClassName);
...
+     */
+    public function evict(string $className): bool
+    {
+        $lowerName  = strtolower($className);
+        $classValue = Core::$executor->classTable->find($lowerName);
+        if ($classValue === null) {
+            return false;
+        }
+
+        $classEntry = $classValue->getRawClass();
```
lisachenko: We don’t need raw value, instead we can use ReflectionClass API, eg isImmutable(), isPreloaded() to make code better and to avoid dealing with class_entry API from outside.

AGENTS.md
lisachenko: Done in b1c0b92 — section dropped. The `@internal` / FFI-CData part is now a paragraph at the end of *Public APIs never leak CData*, saying the same line holds for packages built on z-engine, with the two examples of the fix being a named public method here rather than a reach-through there.

---

src/Core.php
lisachenko: Dropped in b1c0b92 — `init()` and `isInitialized()` only.

Your bootstrap idea removed the need for it entirely rather than relocating it. The consumer no longer asks "can the engine run here?" at all: it requires the autoloader, and code that needs the engine calls the idempotent `Core::init()` — free after a successful auto-boot, and the explanation on a host where that boot was deliberately silent. So the checks stay inside `init()`, exactly as you said.

---

src/Reflection/ClassSpecializer.php
lisachenko: Done in b1c0b92. `evict()` now goes through `ReflectionClass::fromCData()` and reads the class with `isUserDefined()`, `isImmutable()` and `isPreloaded()` — no `ce_flags`, no `->type`, and the lookup reuses the existing private `findClassEntry()`.

`isPreloaded()` did not exist, so it is added next to `isImmutable()` as the same two-line flag accessor. That also removes the `Core::engineConstant('ZEND_ACC_PRELOADED')` string lookup the first version used.

---

PR #188, opened 2026-08-11
feat: typed & IDE-autocompleted engine structs via generated stub classes

lisachenko:
## What & why

Every internal FFI/CData structure is now **statically typed and IDE-autocompleted**. Previously PHPStan saw `mixed` after the first field hop, the IDE autocompleted nothing, and the codebase leaned on ~383 `assert()` calls, 9 hand-written `typeAliases` object shapes, and a large baseline to compensate. This replaces the shape-alias convention with **generated PHP stub classes** produced by the same clang-AST pipeline that emits `engine.h`, plus a templated `Core::new(zend_string::class)` / `cast()` API.

Runtime behaviour changes nowhere — every change is names, docblocks, native-type widening, and generator output. The full suite (`composer test`, plus the destructive `internal` group) stays green and PHPStan is clean at level max.

## Approach

**A — Generated struct stubs.** A new `StructStubEmitter` (`tools/generator/lib/`) derives one analysis-only `final class` per engine struct/union into `stubs/zend-engine-structs.php` (namespace `ZEngine\Generated`, whose short names are the raw C type names). Every C field becomes a real typed public property (scalars → `int`/`float`/`bool`/`string`, pointers → `?otherStub`, arrays/opaque → `\FFI\CData`); named anonymous nested records become synthetic sibling classes (`zval_u1`, `zend_function_common`). The stubs are **never loadable** (`stubs/` is outside the PSR-4 roots), never instantiated, never `instanceof`'d — the only legal runtime use is the `::class` constant. Fields that differ across targets are declared in a new `stub_platform_fields` manifest key and omitted everywhere, keeping the file byte-identical across all 8 platform targets; the `header-drift` CI job now diffs `stubs/` and `.phpstorm.meta.php` alongside `include/`.

**B — Typed entry points.** `Core::new()/cast()/type()/trackedNew()/sizeOfType()/pointerAtAddress()` accept either a C type-name string (unchanged) or a stub `::class`: `new(zval::class)` allocates a `zval`, `cast(zend_string::class, $p)` is a **pointer** cast to `zend_string*`. Static resolution is handled by a small PHPStan `DynamicStaticMethodReturnTypeExtension` (`tools/phpstan/`) plus a generated `.phpstorm.meta.php` for the legacy string literals.

**Migration.** The Type layer, `ReflectionValue`, the function-like reflection core, the class machinery, and the System frame/opline all type their `$pointer` as the owning stub and narrow once at the boundary. Weak `assert($x instanceof CData)` scaffolding is replaced by genuine engine-invariant guards (non-null pointer fields the engine guarantees). The entire `typeAliases` block and the now-dead global `FFI\CData` ignores are deleted; the baseline shrinks from ~663 to 403.

**H — Upstream proposal.** `docs/upstream/` carries a ready-to-post php/php-src issue + a wiki-style RFC draft proposing native opt-in FFI struct→class mapping (targeting PHP 8.6), using this project as the motivating case study.

## The raw-CData escape hatch stays

Per `AGENTS.md`, the pointer-surgery / file-cache zones (`PayloadRelocator`, `ReflectionOpcacheFile`, the persistent-heap cloners) keep the string-form `cast()` and their path-scoped ignores — that is the sanctioned raw-CData boundary, not something the stubs replace.

## Verification

- `composer phpstan` clean at level max (baseline only shrank); `composer test` + `test:internal` green.
- Determinism: `composer gen-headers` (native mode here) run twice → `git diff` clean on `include/`, `stubs/`, `.phpstorm.meta.php`.
- New `tests/CoreTypedEntryPointsTest.php` asserts `sizeOfType(zend_string::class) === sizeOfType('zend_string')`, the class-vs-string cast identity, and that stub classes never load at runtime.
- **Not automatable in CI:** PhpStorm completion (`Core::new(\ZEngine\Generated\zval::class)->` completing `u1`/`value`) — needs a manual check.

## Follow-ups / review notes

- The darwin/windows header workflows need a dispatch run to confirm the stub file is byte-identical on those targets (or to surface more `#ifdef` fields for `stub_platform_fields`); the linux drift job covers linux nts/zts already.
- Namespace `ZEngine\Generated` is fresh (the test analysis-stubs use `ZEngine\Stub`), so no collision.

---
_Generated by [Claude Code](https://claude.ai/code/session_01B8noeLMuYxLRMpTLAinHvL)_


docs/upstream/php-src-ffi-struct-class-map-rfc.md
```diff
@@ -0,0 +1,112 @@
...
+]);
+
+final class Point extends \FFI\CData
+{
+    public function distanceTo(Point $other): float { /* reads $this->x, ... */ }
+}
+```
+
+A registered class must be `final`, must extend `FFI\CData`, and must declare no
+instance properties and no constructor. Given the map, every `CData` ext/ffi
```
lisachenko: Why declare no properties? It can be interesting to have get set hooks defined for fields, operating on raw CData structure

docs/upstream/php-src-ffi-struct-class-map-rfc.md
```diff
@@ -0,0 +1,112 @@
...
+types are minted as instances of the user's class. It is fully backward
+compatible and zero-overhead when unused.
+
+## Proposal
+
+When an FFI scope is created, the author may register a map from C type names to
+classes:
+
+```php
+$ffi = FFI::cdef($code, $lib, classMap: [
```
lisachenko: I would like to propose to use array $options, similar to how SoapServer receives class map binding

docs/upstream/php-src-ffi-struct-class-map-rfc.md
```diff
@@ -0,0 +1,112 @@
...
+- **To ext/ffi internals:** one `HashTable` per scope keyed on `zend_ffi_type *`,
+  consulted only when non-empty at the single `object_init_ex(…,
+  zend_ffi_cdata_ce)` mint site; registration-time validation of the target
+  class. Object layout and handlers are unchanged.
+
+## Open Issues
+
+1. **API surface** — `classMap:` argument vs `#[\FFI\CStruct]` attribute vs
+   `FFI::mapType()`; possibly support more than one.
+2. **Mapping scalars / typedefs** — v1 restricts mapping to struct/union types;
```
lisachenko: Need to cover the type map as well, it can be given as ‘typemap’ array key inside “array $options” configuration.

docs/upstream/php-src-ffi-struct-class-map-rfc.md
```diff
@@ -0,0 +1,112 @@
...
+   whether to later allow scalar typedefs (e.g. a `Handle` wrapper over an `int`)
+   is deferred.
+3. **Unregistered `extends FFI\CData`** — hard error vs inert never-minted class.
+4. **`FFI::cast()` across scopes** — behaviour when casting a handle into a type
+   mapped in a *different* scope.
+
+## Future Scope
+
+- Methods and, potentially, mapped scalar wrappers.
```
lisachenko: Include methods too, there is nothing complex to enable methods on objects

docs/upstream/php-src-ffi-struct-class-map.md
```diff
@@ -0,0 +1,142 @@
...
+
+1. a `classMap:` named argument on `FFI::cdef()`/`FFI::load()` (shown above),
+2. a `#[\FFI\CStruct('name')]` attribute on the class, discovered at scope
+   creation,
+3. a fluent `FFI::mapType('name', Class::class)` call after scope creation.
+
+Happy to prototype whichever the maintainers prefer, or a combination.
+
+---
+_Generated by [Claude Code](https://claude.ai/code)_
```
lisachenko: Do not add Claude Code attribution here please in RFC tickets

stubs/zend-engine-structs.php
```diff
@@ -0,0 +1,3447 @@
...
+ */
+final class zend_arg_info implements \ArrayAccess
+{
+    public ?zend_string $name;
+    public zend_type $type;
+    public ?zend_string $default_value;
+
+    private function __construct() {}
+
+    public function offsetExists(mixed $offset): bool
```
lisachenko: Why do we need everywhere ArrayAccess implementation? If it isn’t array primary, but a struct?

phpstan-baseline.neon
```diff
@@ -889,195 +883,147 @@ parameters:
...
 			count: 1
-			path: src/Reflection/ReflectionClass.php
+			path: src/Memory/PersistentHeap.php
 
 		-
-			message: '#^Binary operation "&" between mixed and 2048 results in an error\.$#'
-			identifier: binaryOp.invalid
-			count: 1
-			path: src/Reflection/ReflectionClass.php
+			message: '#^Parameter \#3 \$pointer of method ZEngine\\Memory\\PersistentHeap\:\:addPointerEntry\(\) expects FFI\\CData, ZEngine\\Generated\\HashTable given\.$#'
```
lisachenko, line 952: Update all signatures to use object typehint instead of CData. And only in  php doc block specify the union type CData + generated object tyie

lisachenko:
Thanks for the review — all seven points are addressed. Pushed in three commits (`72ccc81` docs, `b3b22be` stubs/ArrayAccess, `9000d5d` signatures):

**Docs (RFC + issue)**
- **`array $options` API** — reworked both drafts to a single `$options` array on `FFI::cdef()`/`FFI::load()`, mirroring `SoapServer`/`SoapClient`, instead of the named `classMap:` arg / attribute.
- **`typemap`** — documented a `typemap` key alongside `classmap` inside `$options` for C-type → conversion callbacks (this also covers the scalar/typedef direction I'd previously deferred).
- **Properties/hooks** — dropped the "no properties, no constructor" restriction: a mapped class may now declare typed property hooks over the raw CData…
- **Methods** — …and carry ordinary methods; both are called out explicitly.
- **Attribution** — removed the Claude Code footer from the upstream drafts.

**Code**
- **ArrayAccess on stubs** — you're right, a struct isn't an array. The generator no longer emits `ArrayAccess` on the stubs; `Type\StructArray` is now the single audited pointer-index primitive (it gained a static `at()` raw-deref for the `ptr[0]` case), and every former stub-index site routes through it — array iteration via a `StructArray` instance, single derefs via `StructArray::at()`. Locally-allocated raw `T[N]` CData buffers stay raw.
- **`object` typehints + phpdoc union** — swept ~185 signatures across 43 files: native type is now `object`, with `@param CData` / `@param CData|<stub>` (and `@return`) in the phpdoc only. Two intentional hold-outs: an inline FFI-callback arrow fn and a callback shape described in prose.

PHPStan is clean at level max, `composer test` + the `internal` group are green, and `gen-headers` stays deterministic.

---

lisachenko:
Session:
- Fable 5: 482 in / 1.3k out / 76.4M cache read / 3.3M cache write
- Opus 4.8: 756 in / 2.4k out / 214.4M cache read / 5.2M cache write
- Cost: $39.06 | API 18m | Wall 25m

image

PR #196, opened 2026-08-13
perf(reflection): mint instances via native reflection and drop redundant casts on hot paths

lisachenko:
Two independent, behavior-preserving reductions on reflection hot paths. No public API changes.

> Note: the PR originally carried a third part (entry-level static helpers used by `countPublishedShares()`); it was dropped per maintainer review — consumers hold `Reflection*` objects, entry-level static utilities cut against the framework's model.

## A. Stop building full z-engine `ReflectionClass` instances just to bypass constructors

Inside `namespace ZEngine\Reflection` the unqualified name `ReflectionClass` resolves to z-engine's own class, whose constructor runs the native `parent::__construct()`, a class-table `find()` and `initLowLevelStructures()` (four `HashTable` wrappers) — every bit of it discarded when the instance exists only to serve `newInstanceWithoutConstructor()`.

All ten such sites now use `ReflectionClass as NativeReflectionClass`, the alias already used by `ReflectionClass::fromCData()` (`src/Reflection/ReflectionClass.php:183`) and the `src/Type/*` wrappers:

- `src/Reflection/ReflectionMethod.php` — `fromCData()`, `fromHookCData()`, `fromRawEntry()`, `fromClosureEntry()`
- `src/Reflection/ReflectionProperty.php` — `fromCData()`, `fromRawEntry()`
- `src/Reflection/ReflectionClassConstant.php` — `fromCData()`, `fromRawEntry()`
- `src/Reflection/ReflectionFunction.php` — `fromCData()`, `fromClosureEntry()`

**Deliberately left unchanged:** `src/Reflection/ReflectionMethod.php:152`, `(new ReflectionClass($boardName))->getMethodTable()` — that one genuinely needs the z-engine class, since `getMethodTable()` only exists once `initLowLevelStructures()` has run. Every other site was checked the same way.

## B. Remove redundant `zend_class_entry *` casts

Each of these passed an already-typed `zend_class_entry *` through `Core::cast('zend_class_entry *', …)`, which costs a `count()` array probe, a caught `FFI\Exception`, and a type-name lookup per call. The producing expression was checked at every site — all four are pointer struct fields (`zend_object.ce`, `zend_function_common.scope`), never a C array needing decay — so the cast is dropped, matching the existing direct form in `ReflectionProperty::getDeclaringClass()` and `ReflectionClassConstant::getDeclaringClass()`:

- `src/Type/ObjectEntry.php:131` — `$this->pointer->ce`, `zend_object.ce` is `?zend_class_entry` in the stubs
- `src/Reflection/ReflectionMethod.php:332` — `$this->getCommonPointer()->scope`
- `src/Reflection/FunctionLikeTrait.php:284` — `$entryScope`, fed from `$this->getCommonPointer()->scope` at line 198
- `src/Reflection/FunctionBodySwap.php:406` — `$entryFunction->getCommonPointer()->scope`

No site had to be left in place for array decay.

## Merge note

`ReflectionMethod::fromHookCData()` is also touched by the concurrently-open exception-safety PR #197. The hunks are different and roughly ten lines apart, so a textual proximity conflict is possible but the resolution is trivial in either merge order — keep both edits.

## Validation

- `vendor/bin/phpstan analyse` (level max) — **clean**, no baseline changes needed (verified via a temporary `reportUnmatchedIgnoredErrors: true` run on both trees).
- `PHP_CS_FIXER_IGNORE_ENV=1 vendor/bin/php-cs-fixer fix` — 0 of 296 files changed; `--dry-run` clean.
- **Local tests were not run and could not be:** the container runs PHP 8.5.9 while this branch targets PHP 8.4, and AGENTS.md forbids executing z-engine code against a mismatched minor. Validation here is static-only; CI on the 8.4 legs is the real gate.

### DIFF
--- a/src/Reflection/FunctionBodySwap.php
+++ b/src/Reflection/FunctionBodySwap.php
@@ -222,13 +222,11 @@ final class FunctionBodySwap
             }
             $seenClasses[$classAddress] = true;
 
-            // Both answers come straight off the entry through the owning class: a full
-            // reflection per engine class would pay a native parent constructor and the
-            // low-level table wrappers just to serve one flag read and one table lookup
-            if (!ReflectionClass::entryIsUserDefined($rawClass)) {
+            $reflectionClass = ReflectionClass::fromCData($rawClass);
+            if (!$reflectionClass->isUserDefined()) {
                 continue;
             }
-            $bucketValue = ReflectionClass::entryMethodTable($rawClass)->find($lowerName);
+            $bucketValue = $reflectionClass->getMethodTable()->find($lowerName);
             if ($bucketValue !== null && Core::addressOf($bucketValue->getRawFunction()) === $entryAddress) {
                 $shares++;
             }
--- a/src/Reflection/FunctionBodySwap.php
+++ b/src/Reflection/FunctionBodySwap.php
@@ -222,11 +222,13 @@ final class FunctionBodySwap
             }
             $seenClasses[$classAddress] = true;
 
-            $reflectionClass = ReflectionClass::fromCData($rawClass);
-            if (!$reflectionClass->isUserDefined()) {
+            // Both answers come straight off the entry through the owning class: a full
+            // reflection per engine class would pay a native parent constructor and the
+            // low-level table wrappers just to serve one flag read and one table lookup
+            if (!ReflectionClass::entryIsUserDefined($rawClass)) {
                 continue;
             }
-            $bucketValue = $reflectionClass->getMethodTable()->find($lowerName);
+            $bucketValue = ReflectionClass::entryMethodTable($rawClass)->find($lowerName);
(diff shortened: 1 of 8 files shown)


lisachenko:
Drop B. it is against the idea of framework

lisachenko:
Done — Part B is reverted in 8100b30: the `entryIsUserDefined()`/`entryMethodTable()` statics are removed and `countPublishedShares()` builds `ReflectionClass::fromCData()` per class again, as before. Parts A and C (native reflector for instance minting, redundant cast removal) are unchanged; the PR description is updated accordingly. The concurrently-running hook-consolidation agent was also told not to reuse those statics.

phpstan (level max) and php-cs-fixer re-verified clean after the revert.

---

PR #198, opened 2026-08-13
ci: permissions, concurrency, dependabot coverage, SHA-pinned actions, caches, composer validate/audit

lisachenko:
CI/supply-chain hardening only — no source code is touched. Landing on `8.4` (the minimum affected branch) so it cascades up to `master` through the normal merge-up flow.

**These are workflow changes, so the real verification is this PR's own CI run** — the 14-leg matrix (Linux/macOS/Windows × NTS/ZTS, header-drift, debug-build internal group) has to come back green here before this is trustworthy. Everything below was validated locally only by parsing the YAML (`yaml.safe_load` on all five files) and by running `composer validate --strict --no-check-lock` (passes) and `composer audit` against a freshly resolved lock (no advisories, exit 0).

## Workflow permissions

- `ci.yml` had no `permissions:` block at all, so it inherited the repository default token scope. It now declares top-level `permissions: contents: read`. Every job in it is build/test only — nothing pushes, comments, or writes artifacts back — so no job needed a per-job widening.
- `merge-up.yml` (`contents: read` + `pull-requests: write`) and both `generate-*-headers.yml` (`contents: write`, they commit generated headers) already declared theirs and are unchanged.

## Concurrency

- `ci.yml`: `group: ${{ github.workflow }}-${{ github.ref }}`, `cancel-in-progress: ${{ github.event_name == 'pull_request' }}`. Superseded PR runs are cancelled; pushes to a version branch run to completion so the branch keeps a full result.
- **`generate-darwin-headers.yml` / `generate-windows-headers.yml`: yes, these needed one — and a non-cancelling one.** Both end in a `commit` job that rebases and pushes generated headers to the target branch. Two overlapping runs (e.g. a `workflow_dispatch` refresh while a generator PR run is in flight) race on that push, and cancelling mid-run can leave a matrix leg's artifacts half-refreshed with nothing publishing them. They get `cancel-in-progress: false`, grouped on the branch actually being pushed to (the same `pull_request.head.ref || github.ref` expression the checkout uses), so runs queue instead of racing.
- `merge-up.yml` got the same non-cancelling group: two pushes in quick succession would otherwise both run the "is a merge-up PR already open?" check concurrently and both create one.

## SHA-pinned third-party actions

Pinned to full commit SHAs with the tag they resolve as a trailing comment. `actions/*` (checkout, cache, upload-artifact, download-artifact) intentionally stay on major tags — first-party, and dependabot manages them.

| Action | Was | Pinned SHA | Resolves to |
|---|---|---|---|
| `shivammathur/setup-php` | `@v2` | `f3e473d116dcccaddc5834248c87452386958240` | `v2.37.2` |
| `ramsey/composer-install` | `@v3` (a branch, not a tag) | `a8d0d959dab41457692a5e2041bd9b757a119e3f` | `3.2.1` |
| `ilammy/msvc-dev-cmd` | `@v1` | `0b201ec74fa43914dc39ae48a89fd1d8cb592756` | `v1.13.0` |
| `docker/setup-buildx-action` | `@v3` | `8d2750c68a42422c14e847fe6c8ac0403b4cbd6f` | `v3.12.0` |

Each SHA is the commit the floating ref pointed at when this PR was written (annotated tags peeled to their commit).

## Dependabot coverage

`dependabot.yml` only watched composer. Added:

- `github-actions` (weekly, `/`) — also keeps the new SHA pins and their `# vX.Y.Z` comments updated.
- `docker` (weekly) for `/tools/generator` and `/tools/docker`. Two entries because the docker ecosystem scans one directory each; the fetcher matches any filename containing "dockerfile", so `tools/docker/php-debug.Dockerfile` is picked up despite the non-standard name.
- Caveat worth knowing: both Dockerfiles take their base image through an `ARG` (`BASE_IMAGE=php:8.4-cli`, `FROM php:${PHP_VERSION}-cli`), and dependabot does not rewrite interpolated `FROM` lines. These entries may therefore stay quiet until a literal `FROM` appears — which is also fine, since the PHP minor in those images is branch-pinned on purpose and must not be bumped by a bot.

## composer validate + audit (static-analysis job)

- `composer validate --strict --no-check-lock` before the install step. `--no-check-lock` because `composer.lock` is gitignored (library, nothing pinned) — without it the check fails on the missing lock. Verified locally: passes on composer 2.8.12.
- `composer audit` after the install step. It audits the installed packages, so it must run after `ramsey/composer-install` resolves and installs (there is no committed lock to audit before that). Green today. Note that on composer ≥ 2.8 an *abandoned* package fails this step; if that ever becomes noise for a transitive dev dependency, `composer audit --abandoned=report` is the relief valve.

## Caches

- **PHPStan result cache** — needed a deterministic location, so `phpstan.dist.neon` gains `tmpDir: var/phpstan` under `parameters:`. `.gitignore` already contains `/var/`, so no ignore change was necessary. Cached with `actions/cache@v4` on `var/phpstan`, keyed on `phpstan.dist.neon` + `phpstan-baseline.neon` + `composer.json`, with `${{ github.sha }}` appended so each run stores a fresh (immutable) entry and `restore-keys` fall back to the newest compatible one.
- **php-cs-fixer cache** — `.php-cs-fixer.cache` (already gitignored), keyed on `.php-cs-fixer.dist.php` + `composer.json` with the same fallback shape. php-cs-fixer invalidates the cache itself when its version or rule set changes.

## Optional adds — drop these commit(s) if you'd rather not have them

The third commit (`docs: add a security policy and a pull request template`) is independent of the CI work and can be dropped without affecting anything above:

- **`.github/SECURITY.md`** — states plainly that there is no dedicated security contact yet and asks for GitHub private vulnerability reporting. Most of it is the distinction this library specifically needs: a segfault from calling z-engine APIs, from a version mismatch, or from passing raw `CData`/`@internal` values is an expected failure mode, **not** a vulnerability; a wrong struct layout on a supported build, corruption reachable through the documented PHP-native surface, or a supply-chain issue in the CI pipeline is.
- **`.github/pull_request_template.md`** — mirrors the CONTRIBUTING.md checklist: target the minimum affected version branch, report the exact `php -v` line / thread-safety mode / arch, keep `composer phpstan` and `composer cs:check` green, and regenerate with `composer gen-headers` rather than hand-editing `include/` or `stubs/`. No template existed before, so this PR's own description is hand-written.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,17 @@ on:
       - '[0-9].[0-9]'
   pull_request:
 
+# Build and test only: no job here writes to the repository or its API.
+permissions:
+  contents: read
+
+# One CI run per ref. Superseded pull-request runs are cancelled (the next push
+# re-runs everything anyway); branch runs are left alone so the history of a
+# pushed version branch keeps a complete result.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 # This branch targets PHP 8.4 only - engine structures are version-specific.
 env:
   PHP_MINOR: '8.4'
@@ -40,7 +51,7 @@ jobs:
         # mutates the very classes opcache would publish as immutable. The JIT is
         # off everywhere: it rewrites the executor internals z-engine hooks into
         # (AGENTS.md).
-        uses: shivammathur/setup-php@v2
+        uses: shivammathur/setup-php@f3e473d116dcccaddc5834248c87452386958240 # v2.37.2
         env:
           phpts: ${{ matrix.ts == 'zts' && 'ts' || 'nts' }}
         with:
@@ -60,7 +71,7 @@ jobs:
             || { echo "::error::Zend OPcache is not loaded - the opcache tests would silently skip"; exit 1; }
 
       - name: Install dependencies
-        uses: ramsey/composer-install@v3
+        uses: ramsey/composer-install@a8d0d959dab41457692a5e2041bd9b757a119e3f # v3.2.1
 
       - name: Run test suite
         run: composer test
@@ -107,7 +118,7 @@ jobs:
       # runner process, JIT off everywhere (see the comment there).
       - name: Set up PHP
         if: steps.artifacts.outputs.present == 'true'
-        uses: shivammathur/setup-php@v2
+        uses: shivammathur/setup-php@f3e473d116dcccaddc5834248c87452386958240 # v2.37.2
         env:
           phpts: ${{ matrix.ts == 'zts' && 'ts' || 'nts' }}
         with:
@@ -139,7 +150,7 @@ jobs:
 
       - name: Install dependencies
         if: steps.artifacts.outputs.present == 'true'
-        uses: ramsey/composer-install@v3
+        uses: ramsey/composer-install@a8d0d959dab41457692a5e2041bd9b757a119e3f # v3.2.1
 
       - name: Run test suite
         if: steps.artifacts.outputs.present == 'true'
@@ -191,7 +202,7 @@ jobs:
       # runner process, JIT off everywhere (see the comment there).
       - name: Set up PHP
         if: steps.artifacts.outputs.present == 'true'
-        uses: shivammathur/setup-php@v2
+        uses: shivammathur/setup-php@f3e473d116dcccaddc5834248c87452386958240 # v2.37.2
         env:
           phpts: ${{ matrix.ts == 'zts' && 'ts' || 'nts' }}
         with:
@@ -227,7 +238,7 @@ jobs:
 
       - name: Install dependencies
         if: steps.artifacts.outputs.present == 'true'
-        uses: ramsey/composer-install@v3
+        uses: ramsey/composer-install@a8d0d959dab41457692a5e2041bd9b757a119e3f # v3.2.1
 
       - name: Run test suite
         if: steps.artifacts.outputs.present == 'true'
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -40,7 +51,7 @@ jobs:
         # mutates the very classes opcache would publish as immutable. The JIT is
         # off everywhere: it rewrites the executor internals z-engine hooks into
         # (AGENTS.md).
-        uses: shivammathur/setup-php@v2
+        uses: shivammathur/setup-php@f3e473d116dcccaddc5834248c87452386958240 # v2.37.2
```
lisachenko: Don’t do it, only versions without hashes

.github/workflows/ci.yml
```diff
@@ -60,7 +71,7 @@ jobs:
             || { echo "::error::Zend OPcache is not loaded - the opcache tests would silently skip"; exit 1; }
 
       - name: Install dependencies
-        uses: ramsey/composer-install@v3
+        uses: ramsey/composer-install@a8d0d959dab41457692a5e2041bd9b757a119e3f # v3.2.1
```
lisachenko: Same here, just check that dependabot automates action updates

lisachenko:
Both review comments addressed in b07d9a3: every SHA pin is reverted to the plain version tag it floated on before (`shivammathur/setup-php@v2`, `ramsey/composer-install@v3`, `ilammy/msvc-dev-cmd@v1`, `docker/setup-buildx-action@v3`), across `ci.yml` and both header-generation workflows.

Dependabot automation of action updates is confirmed in place: the `github-actions` ecosystem entry in `.github/dependabot.yml` (weekly, Monday 04:00, `ci` commit prefix) rewrites the `uses:` version tags across all workflows — its comment was updated to drop the now-obsolete SHA-pin wording. All YAML re-parsed clean.

---

PR #203, opened 2026-08-13
refactor(reflection): deduplicate handler installation and flag setters, fix unchecked AST child index

lisachenko:
Eight copy-paste clusters removed across `Reflection\*`, `AbstractSyntaxTree\*` and `Type\*`, plus one real out-of-bounds write fixed. Net **-615 / +522** lines over 17 files; no public signature changed.

Base is `8.4` (the bug in task 5 exists on both lines, so it lands on the minimum affected branch and merges up).

### 1. `installExtensionHandlers()` driven by a map — 120 → 15 lines
`ReflectionClass::installExtensionHandlers()` was 23 repetitions of `if ($this->implementsInterface(X::class)) { $h = parent::getMethod('__y')->getClosure(); $this->setZHandler($h); }`. The interface / magic-method / installer triples now live in one `EXTENSION_HANDLERS` const and a single loop walks it.

Registration **order is unchanged** — it is the const's insertion order, with `ObjectCreateInterface` first (still the mandatory prerequisite, still guarded by the same `\ReflectionException` before the loop). Supporting a new hook interface is now one map line.

### 2. Twenty-one `setXxxHandler()` bodies funnelled — 349 → 195 lines
Each was byte-identical: fetch the handlers block, `new SomeHook($handler, $handlers)`, `install()`, return. **Every public method and its precise hook return type is kept** — that is the API — but the body is one call to a new private
`installObjectHook(class-string<THook> $hookClass, Closure $handler): THook`. The `@template THook of AbstractHook` shape keeps PHPStan resolving each setter to its concrete hook class at level max.

`setCreateObjectHandler()`, `setGetIteratorHandler()` and `setInterfaceGetsImplementedHandler()` keep their own bodies — they write class-entry slots, not the `zend_object_handlers` block.

### 3. `getDefaultProperties()` / `getDefaultStaticMembers()` — 39 → 47 lines (two bodies → one)
Both built a closure-generator over a zval table only to `iterator_to_array()` it back. They now delegate to one private `readZvalTable()` over a bounds-checked `StructArray`. `getDefaultProperties()` keeps its `: iterable` declaration (public API) and both keep returning an index-keyed array.

### 4. Slot-store helpers consolidated on `Type\StructArray`
`storeAdaptationListSlot()` (ReflectionClass), `storePropertyInfoSlot()` (ReflectionClass) and `storePointerSlot()` (ClassSpecializer) were the same one-line write in three places, none of them bound-checked.

Checked the existing `replace()` first: it byte-copies a value *into* the struct a slot points at, which is a different operation from overwriting the pointer the slot itself carries. So `StructArray` gained a sibling, `storePointer()`, and both now share one `assertInBounds()`. Every call site builds the view over the block it just allocated, with the size it allocated it with.

`precedenceExcludeNames()` was verbatim in both classes. Owner is **ReflectionClass** — it owns `zend_class_entry` and its trait adaptations, and `ClassSpecializer` already depends on it (never the reverse), so the reach-through direction is respected. It is `@internal public static` and now returns a `StructArray` sized by the entry's own `num_excludes`, so the four call sites iterate it instead of re-deriving the loop bound.

The *nullable* `properties_info_table` **reads** deliberately stay raw: a nullable pointer array is a shape the element-typed view does not model, and forcing it would have meant weakening `StructArray`'s generics for every other caller.

### 5. Hand-rolled struct-array indexing → bounds-checked access
- `ClassSpecializer::everyReturnIsVerified()` indexed `$opArray->opcodes[$index]` and `[$index - 1]` raw although `$opArray->last` is right there. Now iterates a `StructArray` built from that count, matching `FunctionLikeTrait::getOpCodes()`. The pre-existing `$index === 0` guard still short-circuits before the predecessor lookup.
- `ReflectionProperty::getHook()` indexed the `zend_property_info.hooks` block by hook kind with **no bound at all**; now sized by `Core::ZEND_PROPERTY_HOOK_COUNT`.

Behaviour is identical for every in-range index; an out-of-range one raises `OutOfBoundsException` instead of touching memory past the block.

### 6. AST child access — the fix commit
`getChildren()` / `getChild()` / `replaceChild()` / `removeChild()` each repeated the same four-line bounds-and-cast block, and **none checked `$index < 0`**: `$index >= $totalChildren` lets `-1` through, so the three writers read and overwrote engine memory *in front of* the node's own `zend_ast`. All four now route through one private child-slot accessor built on `StructArray`, which checks both ends. The out-of-range message keeps its previous wording (with the offending index added). 78 → 60 lines for the four accessors, plus 42 lines of shared, documented accessor.

`Node::replaceChild()` and `ListNode::append()` also reached `$node->node` — a **protected** property — on a parameter typed as the `NodeInterface` **interface**; any implementation outside the `Node` hierarchy made that a fatal error.

**BC-preserving choice:** narrowing the parameter to `Node` would be a BC break for anyone implementing `NodeInterface`, so the parameter types are untouched and the reach goes through a shared `Node::rawNodeOf()` guard that throws a descriptive `\InvalidArgumentException` naming the operation and the offending class. Callers that already pass `Node` instances (i.e. everyone today) see no change.

### 7. One shared reverse constant-name lookup
Five classes rendered an engine number as its constant name with three different cache idioms (`static` property, `static $x` local, hand-built loop) and, in all five, an **unfiltered** `getConstants()`. `Type\ConstantNames` now owns the mapping: public constants only (`ReflectionClassConstant::IS_PUBLIC`), integer values only, cached per (class, filter). Miss behaviour stays at the call site, where the two genuinely differ — `ReflectionValue::name()`, `NodeKind::name()` and `OpCode::name()` still throw, `OpLine::typeName()` and `LiveRange::kindName()` still answer `'UNKNOWN'`.

**Latent bug fixed by the filter:** `NodeKind`'s private `AST_SPECIAL_SHIFT` / `AST_IS_LIST_SHIFT` / `AST_NUM_CHILDREN_SHIFT` (6, 7, 8) were flipped into the map next to the real kinds, so `NodeKind::name(6)` answered `'AST_SPECIAL_SHIFT'` instead of reporting an unknown kind. `OpCode::name()` got the same filter. `LiveRange` keeps excluding `KIND_MASK` explicitly (it is a *public* mask, not a kind). Values declared twice still resolve to the last declaration, exactly like the `array_flip()` calls this replaces — the zval type ids reuse their numbers on purpose.

### 8. Flag-toggle triples
Nine copy-pasted visibility setters (`ReflectionMethod`, `ReflectionProperty`, `ReflectionClassConstant`) doing `&= ~ZEND_ACC_PPP_MASK; |= X` in three different engine fields — and ReflectionMethod's three resolved `getCommonPointer()` twice per call. They now come from one `AccessFlagsTrait`; each class only declares *where* its flags live by implementing `replaceAccessFlags()`, which does the read/modify/write in a single pass.

The `if ($on) { $f |= M; } else { $f &= ~M; }` dance appeared seven more times; those are now one-liners over `setAccessFlag()` / `setFunctionFlag()`, the latter private to `FunctionLikeTrait` so plain `ReflectionFunction` gets it too. Verified by reflection that every setter is still a public method of the same class after the trait composition.

## Quality gates

- `vendor/bin/phpstan analyse` — **clean at level max**, nothing added to the baseline.
- `PHP_CS_FIXER_IGNORE_ENV=1 vendor/bin/php-cs-fixer fix --dry-run` — **clean**.
- `phpstan-baseline.neon`: 9 entries pruned and 2 counts lowered, each one freed by a specific change above (the `NodeInterface::$node` property ignores, the `NodeFactory::fromCData()` "mixed given" ignores, the `Core::cast()` counts in `Node`/`ListNode`, the four `array_flip()` ignores and OpLine's two typing ignores). Detailed in the `chore(phpstan)` commit.

## Testing

**Local tests were not possible and were not run.** The container runs PHP 8.5.9 while this branch targets PHP 8.4; per AGENTS.md's non-negotiable version rule, nothing that reaches `Core::init()` may be executed against a mismatched minor. Validation here was static-only (PHPStan + php-cs-fixer) plus two engine-free checks that touch no FFI: loading the changed classes to confirm the trait composition resolves and every public setter survives, and exercising `ConstantNames` through `NodeKind::name()` / `LiveRange::kindName()` (pure `ReflectionClass::getConstants()`, no engine). **CI on 8.4 is the real gate for this PR.**

## Note on #196 / #197

Both concurrently-open PRs touch `ReflectionClass.php` in **different regions** from this one — the `fromCData`-area statics (~line 183 and ~2900) and `addInterfaces` (~286). This PR's edits live in the trait-adaptation helpers (~1100-1500), the properties-info table helpers (~1750-2110) and the whole handler-installation block (~2190-2760); no gratuitous reformatting was done anywhere near their regions. Any conflict should be textual and trivial to resolve.

## Commits

1. `refactor(reflection): read both class default zval tables through one helper`
2. `refactor(reflection): drive extension-handler installation from a map`
3. `refactor(reflection): consolidate struct-array slot writes on StructArray`
4. `refactor(reflection): index engine struct arrays through the bounds-checked view`
5. `fix(ast): bound-check child indexes and guard foreign node implementations` ← **the fix**
6. `refactor(type): share one reverse constant-name lookup, filtered to public constants`
7. `refactor(reflection): extract the flag-toggle and visibility-setter dances`
8. `chore(phpstan): prune the baseline entries freed by the dedup`

---

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -67,6 +67,7 @@ use ZEngine\Core;
 use ZEngine\Generated\zend_class_entry;
 use ZEngine\Generated\zend_class_name;
 use ZEngine\Generated\zval;
+use ZEngine\Hook\AbstractHook;
 use ZEngine\OpCache\SharedMemoryException;
 use ZEngine\Type\ClosureEntry;
 use ZEngine\Type\HashTable;
@@ -2254,6 +2255,42 @@ class ReflectionClass extends NativeReflectionClass
     }
 
     /**
+     * Class-extension hook registry: interface => [magic method that implements it, installer]
+     *
+     * The map is the single source of truth for installExtensionHandlers(): registration
+     * happens in insertion order, and ObjectCreateInterface comes first because the
+     * create_object handler is the mandatory prerequisite of every other one. Supporting a
+     * new hook interface is one line here, nothing else.
+     *
+     * @var array<class-string, array{string, string}>
+     */
+    private const EXTENSION_HANDLERS = [
+        ObjectCreateInterface::class             => ['__init', 'setCreateObjectHandler'],
+        ObjectCastInterface::class               => ['__cast', 'setCastObjectHandler'],
+        ObjectDoOperationInterface::class        => ['__doOperation', 'setDoOperationHandler'],
+        ObjectCompareValuesInterface::class      => ['__compare', 'setCompareValuesHandler'],
+        ObjectReadPropertyInterface::class       => ['__fieldRead', 'setReadPropertyHandler'],
+        ObjectWritePropertyInterface::class      => ['__fieldWrite', 'setWritePropertyHandler'],
+        ObjectGetPropertyPointerInterface::class => ['__fieldPointer', 'setGetPropertyPointerHandler'],
+        ObjectHasPropertyInterface::class        => ['__fieldIsset', 'setHasPropertyHandler'],
+        ObjectUnsetPropertyInterface::class      => ['__fieldUnset', 'setUnsetPropertyHandler'],
+        ObjectGetPropertiesForInterface::class   => ['__getFields', 'setGetPropertiesForHandler'],
+        ObjectGetDebugInfoInterface::class       => ['__getDebugInfo', 'setGetDebugInfoHandler'],
+        ObjectCloneInterface::class              => ['__cloneObject', 'setCloneObjectHandler'],
+        ObjectReadDimensionInterface::class      => ['__dimensionRead', 'setReadDimensionHandler'],
+        ObjectWriteDimensionInterface::class     => ['__dimensionWrite', 'setWriteDimensionHandler'],
+        ObjectHasDimensionInterface::class       => ['__dimensionHas', 'setHasDimensionHandler'],
+        ObjectUnsetDimensionInterface::class     => ['__dimensionUnset', 'setUnsetDimensionHandler'],
+        ObjectCountElementsInterface::class      => ['__count', 'setCountElementsHandler'],
+        ObjectGetClassNameInterface::class       => ['__getClassName', 'setGetClassNameHandler'],
+        ObjectGetConstructorInterface::class     => ['__getConstructor', 'setGetConstructorHandler'],
+        ObjectGetPropertiesInterface::class      => ['__getProperties', 'setGetPropertiesHandler'],
+        ObjectGetClosureInterface::class         => ['__getClosure', 'setGetClosureHandler'],
+        ObjectGetMethodInterface::class          => ['__getMethod', 'setGetMethodHandler'],
+        ObjectGetIteratorInterface::class        => ['__getIterator', 'setGetIteratorHandler'],
+    ];
+
+    /**
      * Installs user-defined object handlers for given class to control extra-features of this class
      */
     public function installExtensionHandlers(): void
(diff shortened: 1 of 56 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -2248,6 +2240,42 @@ public function getReflectionConstant($name)
...
+     * Class-extension hook registry: interface => [magic method that implements it, installer]
+     *
+     * The map is the single source of truth for installExtensionHandlers(): registration
+     * happens in insertion order, and ObjectCreateInterface comes first because the
+     * create_object handler is the mandatory prerequisite of every other one. Supporting a
+     * new hook interface is one line here, nothing else.
+     *
+     * @var array<class-string, array{string, string}>
+     */
+    private const EXTENSION_HANDLERS = [
```
lisachenko: This should be in the beginning of file, not recommended to put constants with methods shuffled

lisachenko:
The `RedefineLeakPlateauTest` failure on this PR is **not a leak** — evidence chain:

1. The failure appeared only once this PR's merge preview included the freshly-advanced `8.4` tip; `8.4`'s own push run at that tip is green, and the failure reproduced byte-identically across two runs and four platforms: `function − evalOnly = 65536` exactly, with `fixedDonor = 0` (pure swap churn perfectly flat — the redefine machinery retains nothing).
2. A temporary diagnostic (b8d3bf9, reverted in 273e2b1) quadrupled the cycles: at 4000 cycles the numbers were `evalOnly=983040, function=1048576, method=1048576, fixedDonor=0` — the overhead **stayed at exactly one 64KB chunk instead of scaling to ~256KB**. A per-cycle leak scales linearly; this didn't.
3. Root cause: both series retain request-arena memory that `memory_get_usage()` observes in whole 64KB ZendMM chunk steps. The measurement cannot resolve sub-chunk differences, and any code-size/allocation-order change in the process (this PR's dedup changes compiled code sizes) shifts intra-chunk alignment enough to deterministically flip one boundary crossing.

Fix in f90398f: `TOLERANCE_BYTES` raised from 16KB to one arena chunk (64KB) with the reasoning documented at the constant. Sensitivity is preserved — a genuine leak ≥ ~66 bytes/cycle still crosses additional chunks at 1000 cycles and fails; sub-chunk alignment noise no longer does. The scaling experiment is reproducible by bumping `$cycles` in `tests/HotSwap/scripts/redefine-plateau.php`.

CI is re-running on the reverted + fixed head.

---

lisachenko:
revert StructArray bound work, it changes a lot of internal details, making this PR complex, let's proceed without it

lisachenko:
StructArray work reverted as requested — commits `4d9f958` (slot-write consolidation), `7e30c26` (bounds-checked indexing in `everyReturnIsVerified`/`getHook`) and `cd0ab27` (AST child access through StructArray) are reverted in place, and the AST baseline stanzas are restored to their `origin/8.4` state. `StructArray` itself, `Node`, `ListNode`, `ClassSpecializer` and `ReflectionProperty` are back to their pre-PR shape.

One thing was deliberately kept, re-implemented minimally (5d1f376): the negative-index defect in `Node::getChild()/replaceChild()/removeChild()` — the three writers accepted `$index < 0` and reached engine memory below the child table. It's now the plain `if ($index < 0 || $index >= $totalChildren)` guard on the existing checks, no structural change. Say the word if you'd rather drop that too.

What remains in the PR: the extension-handler map + `installObjectHook()` funnel, the shared zval-table reader, `ConstantNames` (public-constants filter fixing `NodeKind::name()` on private shift constants), the flag-setter dedup, the const placement fix, and the plateau-tolerance test fix. Title updated accordingly. phpstan level max + php-cs-fixer clean; CI re-running.

---

PR #204, opened 2026-08-13
refactor(memory): route engine-struct access through the owning Type/Reflection classes

lisachenko:
`src/Memory/` was the last corner of the library that reached into engine structures
from the call site: three copies of the same `properties_table` dig, direct
`gc->refcount` / `->properties` / `->handle` / `->ce` pokes, a second array iterator,
two hand-rolled bucket walks and two copies of the same narrowing helpers. This PR
moves every one of them behind the class that owns the structure, as AGENTS.md
requires ("Engine structs are owned by their reflection/type class, never poked from
call sites").

Behaviour is unchanged throughout: same field reads, same write order, same guards,
same exception types.

## 1. Raw zval / zend_object digs → owning API

| Site | Before | After |
|---|---|---|
| `PersistentHeap::attach()` pass 2 | `Core::cast('zval *', ...)` table dig + `u1->v->type` | `getPropertyTablePointer()` + `StructArray` + `ReflectionValue::getBaseType()` |
| `PersistentHeap::attach()` pass 3 | `$obj->ce = ...`, `$obj->properties = null` | `ObjectEntry::setClass()`, `setDynamicPropertiesPointer(null)` |
| `PersistentHeap::evict()` | `$obj->gc->refcount` | `ObjectEntry::getReferenceCount()` (ReferenceCountedTrait) |
| `PersistentHeap::get()` | `newEntry(IS_OBJECT, $root[0])` + materialize | `ObjectEntry::getNativeValue()` |
| `PersistentHeap::releasePropertiesCache()` | `$obj->properties` read/write | `get/setDynamicPropertiesPointer()` |
| `PersistentHeap::currentValidHandle()` | `$obj->handle` | `ObjectEntry::getHandle()` |
| `PersistentHeap` (8 sites) | `$table->getRawValue()->nNumOfElements` | `HashTable::count()` |
| `PersistentGraphCloner::validateObject()` | `ce->ce_flags`, table dig | `ObjectEntry::getClass()->getFlags()`, `getPropertySlot()` |
| `PersistentGraphCloner::cloneObject()` | `ce->name`, `$clone->handle = 0`, table dig | `getClass()->getName()`, `setHandle(0)`, `getPropertySlot()` |
| `PersistentGraphCloner::rewriteValue()` | `zend_value` member + `type_info` writes | `ReflectionValue::setUncountedPayload()` + typed `getRaw*()` |

One deliberate exception: the heap's **re-attachment pass 2 keeps taking the slot count
from the class entry it resolved itself**, not from the object. A stored clone still
carries the class entry of the request that minted it until pass 3 rewrites it, so
`ObjectEntry::getPropertySlotCount()` (which reads the object's own `ce`) would
dereference freed memory there. It reaches the table through the existing `@internal`
`getPropertyTablePointer()` and the blessed `StructArray` view instead — the reason is
spelled out in a comment at the site.

No hot-loop concern: every converted site runs once per object/array at persist or
attach time, never per opcode, so the wrapper allocations are free in practice. No
leaner per-slot accessor was needed.

## 2. HashTable internals → new API surface

**`HashTable::count(): int`** (class now implements `Countable`) — the engine's own
`nNumOfElements`, i.e. the number of entries iteration yields; deleted buckets still
occupying a slot (`nNumUsed`) are not counted.

**`HashTable::findKeyEntry(string $key): ?StringEntry`** *(@internal)* — the bucket's
OWN key block as a borrowed `StringEntry`, which is how a key's storage class is
inspected (interned / permanent / persistent). `null` when the table has no such string
key (packed tables have none by construction).

**`HashTable::replaceKey(string $currentKey, StringEntry $newKey): bool`** *(@internal)* —
swaps that block for an equivalent one. Only the bucket's `key` POINTER is rewritten:
the stored value, the bucket hash and the collision index stay as they are, so the
replacement must carry identical content — enforced with a `LogicException`, since a
different content would silently desynchronise the hash. No refcounting on either side
(the caller owns the new block and must keep it alive at least as long as the table);
returns whether a bucket was actually rewritten.

Both are backed by one private `findBucket()`, now **the only bucket walk in the
framework**. `AbstractModule::makeRegistryKeyPersistent()` is rewritten on the pair and
keeps its exact semantics: a permanent interned key (registered during engine startup)
is left alone, and the persistent interned replacement is only minted when a swap is
actually needed.

### Why no new iterator was added

`PersistentGraphCloner::walkArray()` justified itself with *"unlike
HashTable::getIterator, which drops integer keys of hashed tables"*. That stopped being
true in a56d59d, which taught `getIterator()` to read integer keys from the bucket hash
field (covered by `HashTableTest::testIteratorYieldsIntegerKeysForHashedTables`). The
two changes landed on parallel branches, so the stale duplicate survived the merge.
`getIterator()` is already the key-preserving iteration API for packed and hashed tables
with string and int keys, so `walkArray()` is simply deleted and both call sites
`foreach` over a borrowed `HashTable` view — no `eachEntry()` was added, since it would
only wrap what already composes. **This is the one deviation from the plan**, and it
shrinks the API surface instead of growing it.

## 3. Duplicated narrowing helpers

The character-identical `asCData()` / `asInt()` pairs and the duplicated `zvalType()`
have **no callers left** after (1) and (2), so they are deleted from both files rather
than consolidated into a shared trait. Nothing legitimately remained to host.

## 4. Convention check

The two inline throws inside `PersistentHeap` (`'This persistent heap has been
destroyed'`, `"Corrupt heap metadata: missing entry #{$index}"`) become
`PersistentHeapException::heapDestroyed()` / `::corruptMetadata($index)`, matching their
neighbours (`HeapInertException::create()` etc.). Deliberately scoped to those two — the
codebase-wide sweep belongs to the separate exceptions-convention PR.

## New API summary

| API | Kind | Notes |
|---|---|---|
| `HashTable::count()` | public, `Countable` | live element count |
| `HashTable::findKeyEntry()` | public `@internal` | borrowed key block |
| `HashTable::replaceKey()` | public `@internal` | same-content key swap, no refcounting |
| `ObjectEntry::getPropertySlotCount()` | public | slot count of the object's *current* ce |
| `ReflectionClass::getDefaultPropertiesCount()` | public static | same number from a resolved ce, companion of `getObjectSize()` |
| `ReflectionValue::setUncountedPayload()` | public `@internal` | payload pointer + full `type_info`, no refcounting |

`HashTable::fromCData()` and `PersistentObjectFactory::persistentClone()` additionally
document the `zend_array` / `zend_object` stub views their callers now hold (the engine
declares `HashTable` as a typedef of `struct _zend_array`, so both stub names describe
the same struct).

## Baseline delta

**7 entries removed, none added** (`phpstan-baseline.neon`, −42 lines):

- `src/EngineExtension/AbstractModule.php` — `StringEntry::fromCData() ... mixed given` (the hand-rolled walk lost its `Bucket` type through `Core::addr()`; the walk inside `HashTable` keeps it)
- `src/Memory/PersistentGraphCloner.php` — `cloneObject() expects FFI\CData, zend_object given`
- `src/Memory/PersistentHeap.php` — 3 × `addPointerEntry() expects FFI\CData, ...` (counts 1/5/1)
- `tests/Memory/scenarios/persistent-object-put-detach.php` and `tests/Type/PersistentObjectFactoryTest.php` — `persistentClone() expects FFI\CData, zend_object given`

The remaining `PersistentGraphCloner` entries (inventory properties typed `list<CData>`
vs. stub-typed handles, and the two `@return \FFI\CData` mismatches) are untouched:
widening them only moves the same mismatch onto `PersistedGraph`'s runtime `CData $root`
parameter. They belong to the stub-typing effort, not to this ownership refactor.

## Validation

- `vendor/bin/phpstan analyse` — clean at level max (also verified with
  `reportUnmatchedIgnoredErrors: true` so no baseline entry silently rotted)
- `PHP_CS_FIXER_IGNORE_ENV=1 vendor/bin/php-cs-fixer fix --dry-run` — clean
- `php -l` on every changed file

**Tests were not run locally and could not be**: the container runs PHP 8.5.9 while this
branch targets PHP 8.4, and AGENTS.md forbids executing z-engine (anything reaching
`Core::init()`) against a mismatched minor — the struct offsets differ. CI on 8.4 is the
first real execution. The converted paths are covered there by `PersistentHeapTest`,
`PersistentHeapRequestCycleTest` (its scenarios register the persistent `ZEngineModule`,
which exercises `findKeyEntry()`/`replaceKey()`), `AbstractModuleTest`, `HashTableTest`
and `ObjectEntryTest`.

## Note on adjacency

The concurrently open `sizeOfType` PR touches `PersistentGraphCloner.php` around the
`Core::sizeof(Core::type('HashTable'))` / `Core::type('zval')` lines and
`AbstractModule.php:148`. Those single-line conversions are deliberately left untouched
here; the hunks are adjacent but disjoint, so the two should merge cleanly (the
`$zvalSize` line keeps its exact text and position).

---


src/Reflection/ReflectionClass.php
```diff
@@ -2832,6 +2829,25 @@ public static function getObjectSize(object $classType): int
...
+     *
+     * Companion of getObjectSize() for low-level machinery that holds a resolved
+     * zend_class_entry instead of a live object: the number of zval slots the engine
+     * reserves in every instance (properties_table), the property-guard slot included when
+     * the class uses guards. An object whose class entry is known to be current reads the
+     * same number through ObjectEntry::getPropertySlotCount().
+     *
+     * @param CData|zend_class_entry $classType zend_class_entry to read the slot count of
+     */
+    public static function getDefaultPropertiesCount(object $classType): int
```
lisachenko: instead of returning the default properties count + default properties address - we can use ReflectionClass API and put there all info, instead of operating on raw zend_class_entry

eg ReflectionClass::fromCData($classEntry)->getDefaultProperties()

src/Reflection/ReflectionValue.php
```diff
@@ -593,6 +593,35 @@ public function setPointer(object $pointer): void
...
+     * must either be uninitialized or hold a payload nobody has to release (a byte copy of
+     * another slot, a freshly minted persistent clone).
+     *
+     * <span style="color:red; font-weight:bold">Danger!</span> Low-level API, can bring a segmentation fault
+     * @internal used by the persistent graph cloner
+     *
+     * @param int          $typeInfo Full type_info word: base type | (type flags << Z_TYPE_FLAGS_SHIFT)
+     * @param CData|object $pointer  Payload block; the runtime value is always CData
+     */
+    public function setUncountedPayload(int $typeInfo, object $pointer): void
```
lisachenko: I would like to avoid this method, it is doing low-level manipulation, which is risky, therefore can not be exposed as an API.

src/Type/ObjectEntry.php
```diff
@@ -236,6 +238,25 @@ public function setExtraFlags(int $flags): void
...
+    /**
+     * Returns the number of inline property slots (properties_table) of this object
+     *
+     * The count is a property of the object's CURRENT class entry, so it is only
+     * meaningful while that entry is the live one: machinery that works on an object
+     * whose ce is known to be stale (persistent graphs re-attached in a later request)
+     * must take the count from the class entry it resolved itself, through
+     * ReflectionClass::getDefaultPropertiesCount().
+     */
+    public function getPropertySlotCount(): int
```
lisachenko: this method belongs to the ReflectionClass, do not put it here

lisachenko:
All three review comments addressed in 0ddb381:

- **`ReflectionClass::getDefaultPropertiesCount()` is an instance method now** — no raw `zend_class_entry` crosses the API. The stale-ce pass in `PersistentHeap::attach()` reads it exactly the suggested way: `ReflectionClass::fromCData($resolved[$index])->getDefaultPropertiesCount()`.
- **`ObjectEntry::getPropertySlotCount()` is removed** — the slot count is class information; `ObjectEntry::getPropertySlot()` and the cloner now take the bounds from `$entry->getClass()->getDefaultPropertiesCount()`.
- **`ReflectionValue::setUncountedPayload()` is removed** — the non-refcounted overwrite is only legal on slots the cloner minted itself (byte copies, fresh persistent clones), a precondition no API contract can express. It lives as a `PersistentGraphCloner`-private helper now, reaching the zval through the documented `getRawValue()` escape hatch.

The branch also carries a merge with the current `8.4` tip (including today's #205 hook consolidation) and a cherry-pick of the `RedefineLeakPlateauTest` tolerance fix from #203 (`memory_get_usage()` observes the request arena in 64KB chunk steps; without it this PR's re-run would trip the same deterministic one-chunk alignment flip that #203 hit — full evidence in the #203 thread).

phpstan level max and php-cs-fixer clean; CI re-running.

---

PR #213, opened 2026-08-13
chore(quality): type the test suite, analyse the root files, enforce the license header, add phpstan-phpunit

lisachenko:
Six focused commits paying down analysis debt. Every change is static-only: the container runs PHP 8.5 and this branch targets 8.4, so per AGENTS.md nothing here was executed - no PHPUnit run, no `Core::init()`. Validation is `vendor/bin/phpstan analyse` (level max) and `php-cs-fixer fix --dry-run`, both clean before and after.

**Baseline: 400 entries / 474 errors → 364 / 443.**

## A. `test: type the suite`

Every `missingType.*` entry the baseline carried for `tests/` is gone - 49 of them, across `System/ExecutionDataTest.php` (11 returns + 6 parameters), `Reflection/ReflectionValueTest.php`, `Reflection/ReflectionClassTest.php` (10 returns), `Type/ResourceEntryTest.php`, `Reflection/ReflectionClassConstantTest.php` and `Stub/NativeNumber.php`. Data providers carry precise iterable phpdoc (`list<array{mixed, int}>` and friends) rather than a bare `array`, which is also what lets phpstan-phpunit check provider rows against test signatures.

Typing two stubs properly resolved eight further baselined entries that were pruned as well: `NativeNumber::$value` as `int|float|numeric-string` kills the four `binaryOp.invalid` arithmetic errors, the `getNumericValue()` return type and a `@param` naming a parameter that never existed; `ResourceEntryTest::$file` as a real `resource` kills the `fclose()` argument type and a string cast.

Types only. The single added statement is a `setUp()` guard asserting the fixture stream actually opened, which the `resource` property type requires.

## B. `fix: resolve the dead code and broken phpdoc parked in the baseline`

- **`Compiler::parseString()`** looked like it could never return a tree. `CG(ast)` is cleared right before the parse, so the analyser folded every later `$ast === null` to true and called the return statement unreachable (2× `identical.alwaysTrue` + `deadCode.unreachable`). The field read now goes through a private `getRawAST()` that guards the CData invariant once - the shape `getActiveOpArray()` already uses. `getAST()` and the catch path use it too, which also drops a baselined `argument.type`. Semantics unchanged.
- **`ExecutionData::getCallVariableByNumber()`** carried a `@param CData $call` for a parameter it has not had since the method stopped being static (`parameter.notFound`).
- **`Core::cast()`'s `count()` probe** is the deliberate FFI array-decay trick, exactly as the surrounding comment says: the answer it wants is the `FFI\Exception`, not the number. It keeps the discard and now says so through a scoped `@phpstan-ignore` at the site instead of a baseline line.
- **`Core::preloadFrameworkClasses()`'s `@var`** moves off the iterator (which was never an array) and onto the value it describes.
- **The five `constructor.unusedParameter`** (`ValueNode`, `ReflectionValue`, `ClosureEntry`, `ReferenceEntry`, `StringEntry`) are all false alarms of the same kind, and the most interesting finding here: each constructor reads **its own frame's argument slot 0** (`Core::$executor->getExecutionState()->getArgument(0)`) to capture the caller's zval rather than a copy. The parameter is the mechanism. Removing it would be both a BC break on a public constructor and a functional one. All five keep their signature and now carry that explanation plus a scoped ignore at the site.

Deferred: the `phpDoc.parseError` in `ReflectionClass.php` (`@var (?HashTable|ReflectionValue[])`) - that file belongs to #203.

## C. `chore(quality): analyse and lint the root-level scripts`

`bootstrap.php` runs in every consumer process through composer's `autoload.files` and `preload.php` is the documented `opcache.preload` entry point, yet neither was analysed or linted. The proof was `preload.php`'s `__DIR__.'/vendor/autoload.php'` - unspaced concatenation that @PER-CS2.0 forbids everywhere else in the repository.

Added to the phpstan paths: `bootstrap.php`, `preload.php`, `tools/generator/symbols.php`, `tools/examples/worker-loop.php`. Added to the fixer finder via `->append()`: the two root scripts and the config file itself.

Fallout was small - an alignment fix in `bootstrap.php`, the concatenation plus a missing blank line in `preload.php`, and both scripts plus `symbols.php` are clean at level max. `worker-loop.php` contributes 7 baseline entries, all raw CData dynamics of exactly the kind already baselined for the equivalent `tests/Memory/scenarios` files (hook `proceed()` results, persistent clone handles, array-decay indexes).

## D. `chore(quality): make the license header uniform across the linted corpus`

**This is the one task that did not land as specified, and the numbers are why.** The `$header` the config builds was indeed unused, but wiring it to `header_comment` is blocked on a question this PR cannot answer for you: that fixer enforces one literal header, and the corpus carries three copyright years - **2019 (89 files), 2020 (29), 2026 (168)**. Whichever year the rule hardcodes rewrites the `@copyright` line of every file carrying another:

| enforced year | files rewritten | of which owned by an open PR |
|---|---|---|
| 2026 (minimum churn) | 131 | 14 |
| 2019 (the value in `$header` today) | 210 | 4 |
| 2020 | 270 | 14 |

Every option touches files owned by #199, #201, #203, #204 and #205, and restates when files were written. So the rule is not enabled.

Everything else that was in the way **is** fixed:

- Eleven linted files carried no license header at all, and nine of them opened with a descriptive file docblock that `header_comment` would have silently **replaced** (the generator driver's usage text, the emitters, the manifest, this config). They now carry the header above their own docblock, so the rule would be additive rather than destructive.
- Two files (`tests/System/ObjectStoreTest.php`, `preload.php`) diverged in shape by one line and are normalized.

The corpus is now uniform except for the year, and the config documents the exact four-line rule to switch on once that is settled - verified against the corpus: with it, the only diff the fixer produces is the `@copyright` year itself. The two OpCache payload fixtures are deliberately left alone; their byte content is what the file-cache tests compile and compare.

## E. `chore(quality): teach PHPStan about PHPUnit`

`phpstan/phpstan-phpunit` joins `require-dev` (`composer.json` only - `composer.lock` stays untracked) and its `extension.neon`/`rules.neon` are included **explicitly**, no `extension-installer`, matching how every other service in this config is declared.

Two surfaced smells were real and fixed: an `assertSame(true, ...)` that wanted `assertTrue()`, and a redundant `assertInstanceOf()` on an array element the return type already types (which also freed an import). Two more assertions that only held because the analyser folded a local assignment now read the value back through reflection - which is what the specializer tests meant to check anyway, since the slot is the thing being rewritten.

The `assertCount` suggestion is **declined** for `ReflectionClassDimensionTest` (7 sites, one scoped config ignore with the reason attached): those tests exist to prove a bare `count($object)` reaches the engine's `count_elements` handler on a class that never declared the count itself, and `assertCount()` would measure PHPUnit's `Count` constraint instead of the language construct under test.

The remaining 31 findings are baselined, and they split cleanly in two. Most assert a truth **the engine creates at runtime** and static analysis necessarily cannot see: an interface added to a compiled class, a constant made private, `$this` replaced in the live frame, a runtime-generated specialized class. Those assertions are the suite's whole value and must stay. The rest are type guards the suite keeps as documentation. Three sites whose cause is local and specific carry an inline `@phpstan-ignore` with the reason instead of a baseline line.

One baselined entry (the `assertIsObject` narrowing in `ReflectionClassTest`) is resolved by the extension and pruned. The `assertTrue`/`assertFalse`/`assertNotInstanceOf` entries expected to resolve did **not**: those errors come from PHPStan's core impossible-check rule and survive the extension unchanged.

## F. `docs: merge shadowed docblocks`

19 sites had two consecutive docblocks with nothing between them. PHP attaches only the second, so the first - always the prose explaining what the member is for - was invisible to every IDE and documentation tool while looking perfectly fine in the editor. Each pair is now one block: prose, blank line, tags.

Merged in `src/Reflection/ReflectionValue.php` (12: `getRawClass`, `getRawFunction`, `getRawString`, `getRawArray`, `getRawObject`, `getRawResource`, `getRawReference`, `getRawValue`, `setPointer`, `zvalPointer`, `buildTypeInfo`, `copyAndReleasePrevious`), `src/Core.php` (3: `isTrackedBlock`, `untrackAndFree`, `untrack`) and `src/Type/ResourceEntry.php` (`fromCData`). The three generator entry points hit the same shape from the other direction - the license header added in commit D sat directly above their descriptive file docblock - so their description moves below `declare()`, the shape `symbols.php` and the emitter libraries already use.

## G. Explicitly deferred

`native_function_invocation` plus `strict_param`/`strict_comparison` are **not** in this PR. They rewrite hundreds of lines across `src/` (measured on this branch: 38 backslash-prefixed vs 719 unprefixed native calls) and belong in a dedicated PR once the current modernization wave has merged.

## Deferred sites and adjacency

- **#198** already adds `tmpDir: var/phpstan` to `phpstan.dist.neon`; no `tmpDir` is added here. Both PRs edit that file - #198 in `parameters`, this one in `includes`, `paths`, and `ignoreErrors` - so a merge is mechanical but not textually disjoint.
- **#203** (`ReflectionClass`, `ReflectionMethod`, `ReflectionProperty`, `ReflectionClassConstant`, `FunctionLikeTrait`, `OpLine`, `LiveRange`, `TryCatchElement`, `ArgumentEntry`): skipped the `phpDoc.parseError` entry and 6 shadowed docblocks. `ReflectionValue.php` was touched, but only outside the `name()` cache region.
- **#200** (`src/OpCache/*`, `ClassSpecializer`): skipped 4 shadowed docblocks.
- **#204** (`HashTable`, `ObjectEntry`, `src/Memory/*`): skipped 2 shadowed docblocks (`ObjectEntry.php:289-304` among them).
- **#199** / **#205** / **#201**: no files touched.
- All 12 skipped docblock sites are listed in commit F's message so they are easy to sweep afterwards.

## Not run locally

No tests were executed. The container is PHP 8.5.9 and this branch targets PHP 8.4; per AGENTS.md, running z-engine against a mismatched minor reads the wrong struct offsets. CI on the 8.4 legs is the first real execution of these changes.

---


src/System/Compiler.php
```diff
@@ -180,11 +180,28 @@ public function setCompilationMode(bool $enabled): void
...
+
+    /**
+     * Returns the raw CG(ast) pointer, or null when no tree is attached to the compiler
+     *
+     * The field is only populated while a compilation (or a parseString() run) is in
+     * progress; the engine clears it again when the lexical state is restored.
+     *
+     * @return \FFI\CData|null zend_ast* pointer
+     */
+    private function getRawAST(): ?object
```
lisachenko: do not create such getters around internal data, instead, type $this->pointer with stub php type to allow inference of structure

src/System/Compiler.php
```diff
@@ -357,15 +374,17 @@ public function parseString(string $source, string $fileName = ''): NodeInterfac
         $ast = null;
         try {
             $result = Core::call('zendparse');
-            // restore_lexical_state changes CG(ast) and CG(ast_arena), grab the tree before it
-            $ast = $this->pointer->ast;
+            // restore_lexical_state changes CG(ast) and CG(ast_arena), grab the tree before it.
+            // zendparse() is what fills CG(ast) back in after it was cleared above, so the field
+            // is either the freshly parsed tree or still null when the parser produced nothing
+            $ast = $this->getRawAST();
```
lisachenko: drop method

src/System/Compiler.php
```diff
@@ -357,15 +374,17 @@ public function parseString(string $source, string $fileName = ''): NodeInterfac
...
             if ($result !== Core::SUCCESS) {
                 (new AstOwnership($ast, $arenaBuffer))->release();
                 $ast = null;
             }
         } catch (\Throwable $error) {
             // A ParseError raised by the engine mid-parse: destroy the partial tree and rethrow
-            (new AstOwnership($this->pointer->ast, $arenaBuffer))->release();
+            (new AstOwnership($this->getRawAST(), $arenaBuffer))->release();
```
lisachenko: keep as-is

src/System/ExecutionData.php
```diff
@@ -380,7 +380,6 @@ public function getCallVariable(int $variableOffset): object
      *
      * <span style="color:red; font-weight: bold">Only for the Z-Engine library</span>
      *
-     * @param CData $call zend_execute_data
```
lisachenko, line 399: instead put zend_execute data together with CData for structure awareness with phpstan

lisachenko:
also, resolve conflict with the 8.4 branch

lisachenko:
All review comments addressed in e75c938 + 4c06372, and the branch is conflict-free against the current `8.4` tip:

- **`Compiler::$pointer` is stub-typed** (`@var zend_compiler_globals` over the raw CData handle, boundary narrowing in the constructor) — no getters around internal data; `getRawAST()` is **dropped** and both `parseString()` reads are back exactly as they were.
- The `ExecutionData` frame docblock names the `CData|zend_execute_data` stub union for structure awareness instead of the dropped stale `@param` (the constructor boundary already carried the union).
- The stub typing surfaced real boundaries the CData opacity hid, fixed follow-through: the Core boot narrows the compiler-globals handle, `createArena()` narrows its pointer arithmetic to the `zend_arena` stub, `getAST()` hands `NodeFactory` a raw view, and `setCompilationMode()` writes the bool the stub declares. The CG(ast) null-fold trio (`identical.alwaysTrue` ×2, `deadCode.unreachable`) returned to the baseline — phpstan cannot see `zendparse()` repopulating the field, which is why those entries existed before the rejected getter tried to outsmart them.
- Merge conflicts with `8.4` (post #204/#205/#212) resolved; the baseline conflict resolved by dropping both sides' now-fixed entries (`persistentClone` widening from #204, `cast.string` fixed by this PR's typing).

phpstan level max and php-cs-fixer clean; CI re-running.

---

src/System/Compiler.php
```diff
@@ -184,7 +189,7 @@ public function getAST(): NodeInterface
             throw new \LogicException('Not in compilation process');
         }
 
-        return NodeFactory::fromCData($this->pointer->ast);
+        return NodeFactory::fromCData(Core::cast('zend_ast *', $this->pointer->ast));
```
lisachenko: why cast is required here? check the code

src/Core.php
```diff
@@ -312,8 +312,10 @@ public static function init(): void
             self::$executor = new Executor(self::threadGlobals('zend_executor_globals', $engine->executor_globals_offset));
             self::$compiler = new Compiler(self::threadGlobals('zend_compiler_globals', $engine->compiler_globals_offset));
         } else {
-            self::$executor = new Executor($engine->executor_globals);
-            self::$compiler = new Compiler($engine->compiler_globals);
+            self::$executor  = new Executor($engine->executor_globals);
+            $compilerGlobals = $engine->compiler_globals;
+            assert($compilerGlobals instanceof CData);
```
lisachenko: I don't like this assertion everywhere - please try to drop it or adjust types where needed to avoid assert

lisachenko:
Both follow-ups addressed in b00e5a8:

- **The cast is gone** — you were right, it existed only to feed the analyser. `NodeFactory::fromCData()` and `Node::fromCData()` now declare the `CData|zend_ast` boundary union with a single stub narrowing inside (the same convention as every other `fromCData`), so `Compiler::getAST()` passes `$this->pointer->ast` straight through.
- **The asserts are gone** — the Core boot and `createArena()` narrow the FFI boundary values with `@var` docblocks instead of runtime `assert()` calls; no new asserts remain in the PR.

The baseline follows the widened signature's message drift (two entries updated, one split into its two distinct assignment sites). phpstan level max + php-cs-fixer clean; CI re-running.

---

phpstan-nobaseline.neon
lisachenko, line 1: I would like to have just one single phpstan configuration, not several

lisachenko:
Done in c47bbb7 — `phpstan-nobaseline.neon` is deleted. It was an unreferenced audit convenience (a copy of the dist config minus the baseline include, used to hunt orphaned baseline entries); nothing in composer scripts, CI, or the configs pointed at it. `phpstan.dist.neon` is the single configuration. Gates re-verified clean.

---

lisachenko:
resolve the conflict in phpstan-baseline.neon

lisachenko:
Conflict resolved in 1b51185 — the branch is merged with the current `8.4` tip (post #203). Each of the three baseline conflicts had fixes on both sides: `8.4`'s `ConstantNames` removed the `array_flip` entries while this branch's constructor documentation removed the `constructor.unusedParameter` ones, so the merged baseline drops both. phpstan level max + php-cs-fixer verified clean on the merged tree.

---

PR #217, opened 2026-08-14
refactor: PHP 8.4-era syntax modernization pass

lisachenko:
## What this changes

A pure syntax-modernization pass over `src/` for the PHP 8.4 feature set. No
behaviour, no engine offsets, no public signatures change — every hunk is a
different spelling of the same code.

| Feature | Sites | Where |
|---|---|---|
| `#[\Override]` | **115** methods in 47 files | reflection wrappers, the whole hook layer, `ObjectStore`/`StructArray`/`HashTable`, AST node subclasses, `EngineExtension` modules |
| Typed class constants | **434** constants in 52 files | `OpLine`, `LiveRange`, `ArgumentEntry`, `HashTable`, `ReferenceCountedInterface`, `PersistentObjectFactory`, `NodeKind`, `OpCode`, `PayloadRelocator`, every `HOOK_FIELD`, … |
| `switch` → `match` | **1** | `OpLine::getValuePointer()` |
| First-class callable syntax | **2** | `AbstractHook::install()` / `::refreshTrampoline()` |
| `array_find` / `array_any` / `array_all` | **3** | `ReflectionClass::getMagicSlotFor()`, `HotSwap::assertSourceDeclaresClass()`, `ReflectionMethod::equals()` |
| `str_starts_with` | **2** | `ReflectionExtension::__debugInfo()` |
| `get_class($x)` → `$x::class` | **3** | `IteratorBridge::breakIteration()` (×2), `GetIteratorHook::handle()`, `ReflectionMethod::fromHookCData()` |
| `$x = $x ?? y` → `$x ??= y` | **1** | `BinaryCacheFile::locate()` |

Notes on the judgement calls:

- **`#[\Override]` criterion.** Every candidate was verified programmatically
  against the actual parent-class / interface declaration (PHP itself then
  re-verifies at class-declaration time — a wrong attribute is a fatal error,
  and all 47 files load clean). Constructors are deliberately excluded (LSP
  does not apply to them) and so are trait-provided implementations — a trait
  cannot know whether the using class has a parent to override. This is exactly
  the guard this project wants: it pins one PHP minor and overrides native
  reflection methods whose signatures shift between minors, so drift becomes a
  compile error instead of silent breakage.
- **Nothing became an enum.** ABI-mirrored values stay constants; they only
  gained their type. Array-valued constants are untouched.
- `ReflectionClass::getTraitAliases()` keeps its inline
  `@phpstan-ignore method.childReturnType`; it moved onto the attribute line
  because PHPStan reports the method error at the first attribute line.
- `ReflectionClass::__construct()` keeps `get_class()`: its `$classNameOrObject`
  parameter is untyped, so `::class` cannot be resolved at level max there.
- `PayloadRelocator::(un)serializeZval()` and
  `PersistentGraphCloner::validateValue()` keep their `switch`: the arms are
  multi-statement dispatch, not a value mapping.
- `ClassSpecializer::typeContainsPlaceholder()` /
  `::methodNeedsArgInfoSubstitution()` keep their `foreach`: they iterate
  generators, which `array_any()` does not accept.
- No stray `;` after a `foreach` block exists in `src/`, and the only `!=`
  between same-typed scalars is in a deferred file (below).

### Deferred — sites owned by the still-open #213

These files are left untouched to avoid conflicting with #213; the candidates
are listed so they can be picked up in the follow-up:

- `src/AbstractSyntaxTree/Node.php` — 11 `#[\Override]` sites implementing
  `NodeInterface` (`getKind`, `getAttributes`, `setAttributes`, `getLine`,
  `setLine`, `getChildrenCount`, `getChildren`, `getChild`, `replaceChild`,
  `removeChild`, `dump`); `strpos($methodName, 'get') === 0` at :230 →
  `str_starts_with()`.
- `src/AbstractSyntaxTree/NodeFactory.php` — `switch (true)` at :35 →
  `match (true)`.
- `src/Reflection/ReflectionValue.php` — `switch` → `match` in `equals()` at
  :609; `!=` → `!==` at :871; 25 untyped `int` constants.
- `src/Core.php` — 68 untyped `int`/`string` constants; `substr()` prefix strip
  at :620.
- `src/System/Compiler.php` — 17 untyped constants.
- `tests/` and every config file — untouched, so no test-side `#[\Override]`
  or typed-constant follow-through is included here.

- PHP version (full first line of `php -v`): `PHP 8.5.9 (cli)` — **the container
  runs 8.5, this branch targets 8.4**
- Thread safety: NTS
- OS / architecture: Linux x86_64

**Running the suite locally was not possible** and was not attempted: per
AGENTS.md the one non-negotiable rule is never to run z-engine against a PHP
minor other than the branch's, and `Core::init()` refuses to boot on 8.5
anyway. Verification here is static-only:

- `vendor/bin/phpstan analyse` (level max) — **no errors**
- `PHP_CS_FIXER_IGNORE_ENV=1 vendor/bin/php-cs-fixer fix --dry-run` — **0 of 305
  files need fixing**
- `php -l` over all of `src/`
- every class in `src/` loaded through the autoloader, which makes PHP validate
  all 115 `#[\Override]` attributes and every typed constant at declaration time

Both gates were confirmed green on the pristine `8.4` tip first, so the results
above are attributable to this branch. No `phpstan-baseline.neon` entry was
freed or added by these changes (checked with
`reportUnmatchedIgnoredErrors: true` against both the base and this branch —
the unmatched-inline-ignore count is identical), so the baseline is untouched.
CI on a real 8.4 runner is the authority for `composer test`.

## Checklist

- [x] Targets the **minimum affected version branch** (`master` = newest
      supported minor, `8.4` = PHP 8.4, …) — fixes cascade upward, never
      downward
- [ ] `composer test` passes on the matching PHP minor — **not runnable here
      (8.5 container vs 8.4 target); needs CI**
- [x] Tests added or updated; structs the code dereferences are listed in
      `layout_structs` in `tools/generator/symbols.php` — n/a, no behaviour and
      no new struct dereference
- [x] If `tools/generator/symbols.php` changed: regenerated with
      `composer gen-headers` and committed the result — nothing under
      `include/`, `stubs/` or `.phpstorm.meta.php` was hand-edited — n/a,
      generator and generated artifacts untouched

---


lisachenko:
Resolve conflicts please

lisachenko:
Conflicts resolved in d399ecf — the branch now carries the current `8.4` tip (both #213 and #216 merges). Two genuine overlaps, both between this PR's typed constants and #216's enum migration:

- `ModuleDependency`: kept #216's `@deprecated` int aliases **with** this PR's `const int` types.
- `PersistentHeap`: the `SLOT_*` constants are gone entirely — #216's `DescriptorSlot` enum supersedes them, so this PR's typing of those constants is moot.

Bonus from the unblocking of #213's files: the sweep now also covers `Core.php`, `Compiler.php`, `ReflectionValue.php` and the AST nodes (`#[\Override]` on the `NodeInterface` implementations, the node-factory dispatch as a `match`, remaining string idioms) — the deferred list in the PR body is now empty.

phpstan level max + php-cs-fixer clean on the merged tree; CI re-running.

---

PR #221, opened 2026-08-15
feat(system): consumer-safe file name and compilation-mode bracket for AST hooks

lisachenko:
Closes the reach-through that `lisachenko/scalar-objects` currently needs: it reflects out `Compiler::$pointer` and reads `CG(compiled_filename)` off the raw struct, because no consumer-safe API existed for what its `zend_ast_process` hook must do. Per AGENTS.md — *"when a dependant needs an operation that only exists behind that line, the fix is a named public method here, not a reach-through there"* — this PR adds those named methods.

## The engine behaviour driving this

While `CG(in_compilation)` is set, the engine promotes **every internally-raised exception to a fatal error before any catch runs** — including exceptions thrown and caught entirely inside library code, such as `Core::cast()`'s array-decay probe. Inside a `zend_ast_process` callback that makes:

- nearly every AST accessor fatal unless the flag is cleared around the work, and
- `Compiler::getFileName()` fatal outright (its StringEntry path crosses the throwing probe) — which is a problem, because the per-file gate of an AST consumer must run *before* it decides to touch any state.

## What's added

- **`Compiler::getFileName()`** now reads the `zend_string` bytes directly via `FFI::string()` (element-address decay over the `char[1]` flexible array member) instead of the StringEntry path — no throwing code path, safe inside the hook. Contract otherwise unchanged; an empty compiled filename returns `''`.
- **`Compiler::processInCompilationMode(bool $isEnabled, \Closure $process): mixed`** — enters the requested compilation mode, runs the operation, restores the *previous* mode in a `finally`. Exception-safe in both directions.
- **`AstProcessHook::getFileName()`** and **`AstProcessHook::withoutCompilationMode(\Closure)`** (delegating with `$isEnabled = false`, the direction AST consumers need), so consumers get everything from the hook object they already hold — no `Core::$compiler` (core-layer state), no reflection, no CData.

## Tests

Four new tests in `CompilerTest`:
- `processInCompilationMode` returns the operation result and restores the previous mode in both directions, including when the operation throws;
- `getFileName()` is callable inside a live `zend_ast_process` hook (`internal` group, process-isolated) — **fatals without the fix**;
- the hook's `withoutCompilationMode()` bracket keeps an engine-raised, thrown-and-caught `FFI\Exception` (the exact shape of the cast probe) catchable inside the hook, and restores `in_compilation = true` afterwards.

## Verification

Targeting `8.4` per the branch model (affects both lines; merge-up carries it to `master`). The three changed files are byte-identical on both lines, so the patch was validated on the **8.5 line** locally (this environment runs PHP 8.5.9, and the version-matching rule forbids running the 8.4 line here): full default suite — failure set identical before/after the patch (pre-existing 8.5.9 point-release noise) — plus the internal-group `CompilerTest`, PHPStan level max (no new errors), PER-CS2.0 clean.

Follow-up: once this lands (and merges up), `scalar-objects` drops its reflection/raw-struct code for these two calls (lisachenko/scalar-objects#4).

https://claude.ai/code/session_015M297P4ypTeKaP2u2bfrLq

### DIFF
--- a/src/System/Hook/AstProcessHook.php
+++ b/src/System/Hook/AstProcessHook.php
@@ -16,6 +16,7 @@ namespace ZEngine\System\Hook;
 use FFI\CData;
 use ZEngine\AbstractSyntaxTree\NodeFactory;
 use ZEngine\AbstractSyntaxTree\NodeInterface;
+use ZEngine\Core;
 use ZEngine\Hook\AbstractHook;
 
 /**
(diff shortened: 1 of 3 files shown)


src/System/Hook/AstProcessHook.php
```diff
@@ -56,6 +57,37 @@ public function getAST(): NodeInterface
...
+     * accessor crosses. Consumers therefore run their tree work through this bracket:
+     *
+     *     Core::setASTProcessHandler(function (AstProcessHook $hook): void {
+     *         $hook->withoutCompilationMode(fn () => rewrite($hook->getAST()));
+     *         if ($hook->hasOriginalHandler()) {
+     *             $hook->proceed();
+     *         }
+     *     });
+     */
+    public function withoutCompilationMode(\Closure $operation): mixed
```
lisachenko: Drop this method here, it just delegates and increases complexity. just put info into the doc block of AST hook how to manipulate compiler mode

PR #224, opened 2026-08-15
feat(memory): allocator seam for persistent cloning + external arData install (closes #223)

### LINKED ISSUE #223: Allocator seam for persistent graph cloning + external arData install API (PR-Z1)
Part of the **zero-serialization PHP coroutines** initiative: forked workers exchange real shared PHP objects living in a fork-shared mmap arena (built on `lisachenko/php-shared-data-extension`), with no `serialize()` anywhere on the data path. Full plan: https://claude.ai/code/artifact/52145286-d84d-4df2-a860-1d39f7d930f2

Today all persistent-graph allocation funnels through exactly three primitives, each hardcoding `Core::trackedNew → FFI::new → pemalloc` (process-local malloc), which cannot be redirected:

- `PersistentObjectFactory::persistentClone()` (src/Type/PersistentObjectFactory.php:69)
- `StringEntry::persistent()` / `persistentInterned()` (src/Type/StringEntry.php:134/172)
- `PersistentHashTable` construction (src/Type/HashTable.php:114) — plus `arData` grown by engine C via `pemalloc` with no hook

## Scope

1. **Allocator seam.** Introduce a minimal `Allocator` interface (allocate aligned block, report ownership) and thread an optional allocator parameter through `persistentClone`, `PersistentGraphCloner`, `StringEntry::persistent(Interned)`, and the `PersistentHashTable` constructor. Default = current tracked-pemalloc behavior, byte-for-byte — zero behavior change for existing callers.
2. **External arData install API.** A public method on `PersistentHashTable` to install a **pre-sized, externally allocated** `arData` block post-construction (packed + hashed layout), with growth forbidden/detected afterwards. Today there is no API for this; it is required so shared tables never let the engine grow storage into process-local heap.
3. **Public-API remedies.** Named public methods for the operations `php-shared-data-extension` currently performs via `@internal` / `Core::$executor` reach-throughs (object-store put/recycle, class-table find, engine calls used by PersistentStore/Persister) — per AGENTS.md “a named public method here, not a reach-through there”.

## Constraints

- Lands on the **8.4 branch** first, merges up to master (AGENTS.md branch flow). No cherry-picks down.
- **Generated headers untouched** — mmap and pthread primitives are bound package-locally in the consumer via `FFI::cdef` (precedent: `Core::persistentFree()` Windows cdef, src/Core.php:874).
- No public API may leak `CData` (AGENTS.md).
- PHPStan level max, PER-CS2.0 clean; existing tests green.

## Acceptance criteria

- [ ] `Allocator` interface + default implementation preserving current behavior; all three primitives accept it
- [ ] `PersistentHashTable::installExternalStorage(...)` (naming per style) with growth-guard semantics + tests
- [ ] Reach-through remediation methods with tests
- [ ] Suite green on PHP 8.4 and 8.5 lines per branch flow

Downstream consumer: `php-shared-data-extension` Arena (fork-shared mmap) — see the EPIC in that repository.

lisachenko:
## What this changes

Closes #223 (PR-Z1 of the zero-serialization PHP coroutines initiative).

Today every persistent byte z-engine mints funnels through a hardcoded `Core::trackedNew(..., persistent)` — FFI's `pemalloc`, i.e. the **process-local** heap. A consumer that keeps real PHP objects in a **fork-shared mmap arena** (`lisachenko/php-shared-data-extension`, EPIC lisachenko/php-shared-data-extension#15) cannot redirect that, and a table the engine grows on its own drags half of itself back into process-local memory. This PR opens both seams and removes the reach-throughs that consumer needed.

**1. Allocator seam.** `ZEngine\Memory\Allocator` hands out **addresses** of zeroed, aligned blocks and reports whether it keeps ownership of them. `EngineAllocator` is the default implementation and reproduces the three allocation shapes the framework used to hardcode — `trackedPersistent()`, `trackedRequest()`, `persistent()` (untracked). An optional allocator is threaded through:

- `PersistentObjectFactory::persistentClone($object, ?Allocator)`
- `StringEntry::persistent()` / `persistentInterned()`
- `HashTable::__construct(?Allocator)` (inherited by `PersistentHashTable`)
- `PersistentGraphCloner::__construct(?Allocator)` — one allocator for a whole graph

Null keeps **each primitive on its own historical default**, which is deliberately not the same allocator for all of them (object clones and table structs stay tracked malloc blocks, string blocks stay untracked ones). Zero behavior change for existing callers; no existing test needed a line changed.

**2. External arData install.** `PersistentHashTable::installExternalStorage($address, $capacity)` writes the engine's own `zend_hash_real_init_mixed` layout over a caller-allocated block — `HT_SIZE_EX` sizing via `externalStorageSize()`, hash part of two `uint32_t` slots per bucket reset to `HT_INVALID_IDX`, buckets right behind it, `HASH_FLAG_UNINITIALIZED` replaced by `HASH_FLAG_STATIC_KEYS` while the iterator count is left alone. Installation is only legal **before the first insert**, so the engine never real-initializes storage of its own. `withExternalStorage()` pairs it with the struct allocator in one call.

Growth is guarded from both sides, since the engine grows a full table by `perealloc`ing exactly that block:

- **prevention**: inserts through the wrapper refuse the write that would trigger the resize (`getRemainingCapacity()` reports the headroom; an upsert of an existing key consumes no slot and stays allowed at full capacity);
- **detection**: `assertNoGrowth()` verifies `arData`/capacity are still the installed ones — for engine paths this class cannot intercept — and runs after every insert.

`destroy()` refuses on both foreign-allocated structs and installed external storage: `zend_hash_destroy()` would `pefree` the caller's block and the struct free assumes z-engine's allocator.

**3. Public-API remedies** for the `@internal` / `Core::$executor` reach-throughs in `php-shared-data-extension`:

| Reach-through today | Named public method |
|---|---|
| `Core::$executor->objectStore->put($cdata)` | `ObjectEntry::register(): int` |
| `Core::$executor->objectStore->recycle($handle)` | `ObjectEntry::unregister(): void` |
| `Core::$executor->classTable->find(strtolower($n))` | `ReflectionClass::fromClassTable(string): ?ReflectionClass` (no autoload, null on miss) |
| `Core::type($t)->getStructFieldOffset($f)` | `Core::offsetOfField(string $type, string $field): int` |
| `Core::sizeof(Core::type('zval'))` | already covered by `Core::sizeOfType()` |

Store registration is encapsulated on the object that is being registered (review feedback, f8a4ab5): `register()` returns the handle the engine assigned, `unregister()` reads the handle at call time, verifies the slot actually holds this object and returns it to the free list — a stale handle is refused with a named domain exception instead of detaching somebody else's object. `ObjectStore` itself is unchanged by this PR; `put()`/`recycle()` stay `@internal` and a consumer never sees `EG(objects_store)` at all.

## Design notes

- **No CData crosses a public boundary.** The allocator interface is integers only (`allocate(int $size, int $alignment = 16): int`), so an arena implementation in a consumer package binds its own `mmap`/`pthread` primitives with `FFI::cdef` (precedent: `Core::persistentFree()`'s Windows cdef) and returns the address it computed. `installExternalStorage()` takes an address; `ObjectEntry::register()`/`unregister()` take and return nothing but an `int` handle. Blocks are converted to typed handles inside the owning classes.
- **Ownership is explicit, not guessed.** `Allocator::ownsAllocations()` is what a structure consults before any release path; a table on foreign memory throws `TypeOperationException` instead of calling `free(3)` on somebody else's arena.
- **Engine structures stay owned by their type class.** All `HT_SIZE_EX`/`HT_SIZE_TO_MASK`/`HT_HASH_RESET` arithmetic lives inside `PersistentHashTable`, and the store bookkeeping of an object lives on `ObjectEntry`; callers pass capacities and handles, never offsets or store views.
- **Generated headers untouched** — no `include/`, `stubs/` or `.phpstorm.meta.php` change; no new engine symbol was needed (`HASH_FLAG_STATIC_KEYS`, `HT_MIN_SIZE` and friends are already generated constants).
- **Domain exceptions via named constructors** — eight new factories on `TypeOperationException`, plus a new `AllocationException`.
- Not in scope: `PersistentHeap::put()` still uses the default allocator. Its eviction path frees blocks through z-engine's own allocator, so accepting a foreign one there needs a matching release contract — a follow-up, not a silent half-measure.

## Test evidence

```
composer test                 497 tests, 5095 assertions — OK (5 skipped, 5 incomplete)   [462 before]
phpunit --group internal      161 tests,  480 assertions — OK (process-isolated, no crash)
composer phpstan              [OK] No errors  (level max, no new baseline entries)
composer cs:check             [OK]            (PER-CS2.0)
```

New coverage: `tests/Memory/EngineAllocatorTest.php` (zeroing, alignment guarantees per mode, registry membership, refusals), `tests/Memory/AllocatorSeamTest.php` (clones/strings/tables/whole cyclic graphs land in an arena-shaped test allocator; defaults keep the historical allocation classes), `tests/Type/PersistentHashTableExternalStorageTest.php` (layout math, lookups through installed storage, capacity guard, upsert at full capacity, relocation diagnosis, install-after-init and destroy refusals, struct+storage both in one region), `tests/Type/ObjectEntryRegistrationTest.php` (handle assignment and engine visibility via `spl_object_id`, slot returned without touching the object, re-registration, both refusal paths), plus `fromClassTable()`/`offsetOfField()` cases in the existing Reflection/Core tests.

- PHP version (full first line of `php -v`): `PHP 8.4.19 (cli) (built: Mar 30 2026 19:28:35) (NTS)`
- Thread safety: NTS
- OS / architecture: Linux / x86_64

## Checklist

- [x] Targets the **minimum affected version branch** (`8.4`) — merges up to `master` per branch flow
- [x] `composer test` passes on the matching PHP minor
- [x] Tests added or updated; no new struct dereference needed (`HashTable`/`Bucket` are already in `layout_structs`)
- [x] `tools/generator/symbols.php` unchanged — nothing under `include/`, `stubs/` or `.phpstorm.meta.php` was touched

https://claude.ai/code/session_01B9xaBchjdo1atarNZ6sZqe


src/System/ObjectStore.php
```diff
@@ -109,6 +109,19 @@ public function offsetUnset($offset): void
...
 
+    /**
+     * Returns the object store of the CURRENT request
+     *
+     * The public entry point into EG(objects_store): the engine-global wrappers behind
+     * Core::$executor are core-layer state and not a consumer API (AGENTS.md), so a package
+     * that has to hand the engine an object of its own asks here instead of reaching through
+     * them. The returned view is the one the executor itself uses - stores are never copies.
+     */
+    public static function current(): self
```
lisachenko: It should not be exposed as a separate method. Public access to private properties should be avoided. Instead encapsulate access to internal structures into relevant classes and methods.

lisachenko: Agreed — `ObjectStore::current()` is gone, together with the `register()`/`unregister()` pair that only existed to be reached through it. `src/System/ObjectStore.php` is now byte-for-byte what it was before this PR, `put()`/`recycle()` included, and both stay `@internal`.

The operations moved onto the object they concern, in f8a4ab5:

- `ObjectEntry::register(): int` — registers THIS object in the current request's store and returns the handle the engine assigned (`== spl_object_id`).
- `ObjectEntry::unregister(): void` — reads the object's handle at call time, verifies the slot actually holds this object, and returns it to the free list. A stale handle (never registered, already unregistered, or a slot meanwhile reused) is refused with `TypeOperationException::objectNotRegistered()` instead of silently detaching somebody else's object.

So a consumer never sees `EG(objects_store)` at all — which is a stronger remedy than the original one, since even the store view stays inside the framework. Tests moved accordingly to `tests/Type/ObjectEntryRegistrationTest.php` (fresh handle + engine visibility via `spl_object_id`, slot returned without touching the object, re-registration across detaches, and both refusal paths).

---

Issue #241, opened 2026-08-19
feat(opcache): decline inheritance-cache publication for classes that receive handlers during linking


lisachenko:
## Summary

The real fix for #238. Under opcache, a class linked at runtime against a cached interface is linked on a temporary mutable copy (`zend_lazy_class_load()`), and after linking opcache's **inheritance cache** persists the linked result into shared memory and discards the temporary. Handlers installed from an `interface_gets_implemented` hook are keyed to the discarded temporary's address and are silently lost (full diagnosis in #238's comments).

The interim measure (landing with the opcache CI/test effort) is a loud guard: handler installation on a lazy-linking temporary throws `SharedMemoryException`. This ticket is the follow-up that makes the feature actually **work** under opcache.

## Design

Intercept **`zend_inheritance_cache_add`** — a `ZEND_API extern` function pointer, so it is hookable from FFI — and **decline caching** for classes that received z-engine handlers during linking:

- The linked class entry then stays process-local and mutable; the address-keyed handler table stays valid; handlers fire.
- The multi-worker hazard disappears by construction: no process-local libffi trampoline address is ever published into shared memory / the inheritance cache, so no sibling FPM worker can reach a trampoline that does not exist in its address space.
- Cost: the affected class loses inheritance-cache reuse (re-linked per process/request). Only classes actually hooked during linking pay it.

### Work items

1. Add `zend_inheritance_cache_add` (and any companion symbols needed for a clean decline) to `tools/generator/symbols.php`; regenerate `include/` for all platform targets — Linux via `composer gen-headers` (docker), macOS/Windows via the `generate-darwin-headers.yml` / `generate-windows-headers.yml` dispatch workflows.
2. Wrap the function pointer during `Core::init()`: delegate to the original unless the class (or one of the classes linked in the same request) is registered in the handler table; in that case return the "not cached" result the engine expects.
3. Un-throw the interim guard: with the interception active, handler installation from `interface_gets_implemented` becomes supported; the guard remains only as a fallback when the interception is unavailable (e.g. platform target without regenerated headers).
4. Regression tests: the #238 reproducer (cached interface + hook + implementor, `opcache.file_update_protection=0`) asserting `write_property` fires with opcache enabled; a matrix leg asserting an un-hooked sibling class still lands in the inheritance cache.
5. Docs: update the hot-swap opcache support matrix row added by the interim guard.

## Alternatives considered and rejected (recorded from the #238 discussion)

[...]

Issue #238, opened 2026-08-19
bug(opcache): a write_property handler installed from an interface_gets_implemented hook never fires when opcache is enabled


lisachenko:
## Summary

When handlers are installed on an implementor **from inside an `interface_gets_implemented` hook**, the `write_property` handler silently never fires if opcache is enabled. The interface hook itself fires, and the `create_object` handler installed in the same callback fires — only `write_property` is lost.

There is no error, warning or crash: property writes just take effect as if no handler existed. Any package that gates writes this way (e.g. [`lisachenko/immutable-object`](https://github.com/lisachenko/immutable-object)) silently stops enforcing anything.

Reproduced on **both supported lines**: PHP 8.4.19 + z-engine 8.4.2 and PHP 8.5.9 + z-engine 8.5.0.

## Observed behaviour

`opcache.jit=off`, `ffi.enable=1`, varying only `opcache.enable_cli`:

| PHP / z-engine | opcache | link path | interface hook | `create_object` | `write_property` |
|---|---|---|---|---|---|
| 8.4.19 / 8.4.2 | off | include | fired | fired | fired |
| 8.4.19 / 8.4.2 | off | autoload | fired | fired | fired |
| 8.4.19 / 8.4.2 | **on** | include | fired | fired | **NOT fired** |
| 8.4.19 / 8.4.2 | **on** | autoload | fired | fired | **NOT fired** |
| 8.5.9 / 8.5.0 | off | include | fired | fired | fired |
| 8.5.9 / 8.5.0 | off | autoload | fired | fired | fired |
| 8.5.9 / 8.5.0 | **on** | include | fired | fired | **NOT fired** |
| 8.5.9 / 8.5.0 | **on** | autoload | fired | fired | **NOT fired** |

The link path (plain `include` vs autoloader callback) makes no difference; `opcache.enable_cli` is the only variable that changes the outcome.

## Scope: only the interface-hook path is affected

Installing the same two handlers **directly on an already-declared class**, with no interface hook involved, works under opcache on both minors:

```
php8.4 opcache=1  direct install | create_object: fired | write_property: fired
php8.5 opcache=1  direct install | create_object: fired | write_property: fired
```

That is consistent with the sibling packages being unaffected — `native-php-matrix`, `scalar-objects` and `native-php-coroutines` install their handlers on named classes and their suites pass with opcache enabled.

## Reproducer

```php
<?php
declare(strict_types=1);

[...]

Issue #131, opened 2026-08-05
feat(reflection): copy literals alongside opcodes to lift the 2GB reach limit on a patched RECV


lisachenko:
Follow-up to #129, which is where this restriction was introduced and documented.

## The restriction

Patching the type mask a `ZEND_RECV` caches in its opline means un-sharing the method's opcode array. The opcodes are copied into request memory while the **literals stay shared**, so every `IS_CONST` operand is rebased by the distance moved.

An `IS_CONST` operand stores a **signed 32-bit** offset from its own opline. Request memory and the compiler arena are neighbours, so an ordinary class is comfortably in range — but an **opcache-shared body** lives in an mmap'd region that can be arbitrarily far away. `duplicateOpcodes()` detects that and throws rather than silently truncating the offset:

```
Cannot specialize ...: the relocated opcode array cannot reach its literals
```

## Consequence today

A `mixed` (or otherwise builtin) **parameter** cannot be re-typed when the template's body is in opcache shared memory. Nothing else is affected: class-typed parameters, every return type and every property are untouched, because none of them un-shares an opcode array.

Downstream this surfaces as a documented `markTestSkipped` in `lisachenko/userland-php-generics` (`AttributeFormTest::testAMixedParameterIsRetypedAndEnforced`), which is a fair workaround but leaves a real capability gap under the configuration people actually deploy.

## The fix

Copy the literals along with the opcodes, so the copy owns a contiguous block of its own and every `IS_CONST` offset stays short. The open question is **zval ownership**: the literals table holds refcounted zvals that `destroy_op_array()` releases when the shared body refcount reaches zero. A duplicated table needs its own refcount handling that does not double-release the originals — the same problem `duplicateArgInfo()` already solves for arg names and types, so the existing ownership note is the model to follow.

## Verification

- The existing opcode-relocation test must still pass, with the assert-verification under `zend.assertions=1` extended to check that literals resolve on the copy too.
- `ClassSpecializerShmTest`-style coverage: an immutable, opcache-shared template with a `mixed` parameter must specialize successfully instead of being rejected.
- Downstream, the skip in `userland-php-generics` should be removable, and `composer test` there should stay green under `opcache.enable_cli=1`.

lisachenko:
[...]

PR #244, opened 2026-08-19
feat(opcache): hold the suite under an opcache-active runner — CI job, lazy-linking guard, static-vars and introspection fixes

### LINKED ISSUE #239: bug(opcache): getStaticVariables() reads the defaults table instead of the live static_variables_ptr table for opcache-shared functions
## Summary

For a function whose op_array lives in opcache shared memory (`ZEND_ACC_IMMUTABLE`), `getStaticVariables()` (`src/Reflection/FunctionLikeTrait.php`) returns the **declaration defaults** instead of the live, materialized static-variables table. Static-variable state changes made by calling the function are invisible to the reader.

Found by running the full suite with `opcache.enable_cli=1` in the runner process: `tests/Reflection/FunctionLikeInfoTest::testStaticVariablesOfNamedFunction` fails with `Failed asserting that 0 is identical to 2` — the test calls the function twice and expects the live `invocations` counter, but gets the default `0`. Reproduced identically on the `8.4` line (PHP 8.4.19) and `master` (PHP 8.5.9); deterministic.

## Mechanism

`getStaticVariables()` reads `zend_op_array.static_variables_ptr` and, when the value has the low bit set, falls back to the defaults table (the docblock even documents this as intended). But for immutable (shared-memory) op_arrays `zend_persist.c` stores the **offset form** of `ZEND_MAP_PTR` there: an odd byte offset into the per-process map-ptr area. The live table (materialized by the first `ZEND_BIND_STATIC`) sits in the slot at `CG(map_ptr_base) + offset` and is never dereferenced.

The codebase already resolves exactly this encoding elsewhere: `StringEntry::setCachedClassEntry()` computes the slot as `base + offset` (base is pre-biased, no `-1`) and bound-checks with `(offset - 1) / sizeof(void*) < CG(map_ptr_last)`, mirroring `ZEND_MAP_PTR_OFFSET2PTR` / `ZSTR_VALID_CE_CACHE`.

## Fix shape

In `getStaticVariables()`: when the low bit is set, dereference the map-ptr slot (`Compiler::getMapPointerBaseAddress()` + offset, with the bound check); return the materialized table when the slot is non-NULL, and fall back to the defaults only while the slot is still NULL (before the first call — semantically correct, since the defaults are what the first bind installs). Rewrite the misleading docblock.

`getStaticVariables()` is the only static-variables reader in the codebase, so no other call site is affected.

## Regression coverage

- `FunctionLikeInfoTest::testStaticVariablesOfNamedFunction` as-is, under an opcache-active runner.
- A deterministic leg in the `OpcacheSupportMatrixTest` child script asserting the raw field is in offset form and that the reader sees the live count, so the branch is exercised in the default (opcache-off) CI too.

Applies to branch `8.4` first, cascades to `master` (encoding identical on both lines).

### LINKED ISSUE #240: feat(reflection): optimizer-aware frame-variable presence/enumeration API on ExecutionData
## Summary

With opcache active, the optimizer (SCCP + dead-code elimination + CV compaction) legitimately removes and renumbers compiled-variable slots. Frame introspection then sees the optimized truth: a variable that exists in the source may have **no CV slot at runtime**, and `ExecutionData::getLocalVariable()` throws `OutOfBoundsException` for it.

Found by running the full suite with `opcache.enable_cli=1` in the runner: `tests/System/ExecutionDataTest::testGetLocalVariableObservesUndefSlot` (never-assigned CV dropped after range inference proves its branch dead) and `::testGetLocalVariablesOfParentFrame` (a marker CV removed by SCCP + DCE) fail; both pass with `opcache.optimization_level=0`. Identical on the `8.4` and `master` lines; deterministic.

## Agreed design (discussed with the maintainer)

Hybrid stance:

1. **Document the optimizer reality** — introspection reports the optimized op_array truthfully; instrumentation scenarios that must see every source variable should run with `opcache.optimization_level=0` (`docs/self-debugging.md`).
2. **Keep the throw** on access to a missing variable (`getLocalVariable()` keeps its `OutOfBoundsException`).
3. **Add presence/enumeration API** so clients can probe before access, following the existing naming family:
   - `ExecutionData::hasLocalVariable(string $variableName): bool`
   - promote the existing private `ExecutionData::getLocalVariableNames()` (slot-keyed `array<int, string>`) to public.

Tests adapt to be branch-aware: probe presence first; when the slot is present (default opcache-off CI) the original strict assertions run unchanged, when absent (opcache-active runner) the throw / empty-shape contract is asserted instead.

Applies to branch `8.4` first, cascades to `master`.

lisachenko:
## What this changes

Part of the opcache support meta-epic (bug-collection round): the full suite now holds with **opcache active in the runner process itself** (`opcache.enable_cli=1`, every test file compiled into shared memory) — the consumer-facing environment — with a dedicated Linux CI job gating it.

Running the suite that way on both lines (PHP 8.4.19 / branch `8.4` and PHP 8.5.9 / `master`) produced an identical, deterministic 6-test failure diff; this PR fixes or triages every one of them:

- **Fixes #239** — `getStaticVariables()` now dereferences the map-ptr offset slot opcache stores into shared immutable op_arrays and returns the live table (was: declaration defaults). New opcache-matrix leg proves the branch deterministically.
- **Fixes #240** — `ExecutionData::hasLocalVariable()` + public `getLocalVariableNames()`: probe/enumeration surface for frames whose CV slots the opcache optimizer removed or renumbered; `getLocalVariable()` keeps its documented throw. The two `ExecutionDataTest` cases are branch-aware (strict shape when the slot exists, throw/empty shape when optimized away).
- **Refs #238** — the loud guard agreed in the issue: handler installation on the opcache **lazy-linking temporary** (`ZEND_ACC_CACHED` set, `ZEND_ACC_LINKED` clear — the state an `interface_gets_implemented` hook observes for a cached implementor) throws `SharedMemoryException::handlerInstallationDuringLazyLinking()` instead of being silently lost with the discarded temporary. New public `ReflectionClass::isLazyLinkingCopy()` for probing. Child-process reproduction test (with `opcache.file_update_protection=0` — the default 2s silently degrades it) plus an opcache-off companion proving the same installation fires handlers for real. #241 tracks the actual fix (declining the inheritance cache for hooked classes).
- **Refs #131** — the two `ClassSpecializerSlotTest` methods hitting the 2GB IS_CONST reach refusal join the new issue-linked `opcache-incompatible` group with a `TODO(#131)`; they return to the job when literals travel with opcodes.
- `SharedMemoryExceptionTest` premise fix: under an opcache-active runner the test file's own classes ARE immutable — the runtime-declared probe is now eval-declared.

CI: new `tests-opcache-runner` job (Linux only by design — Windows opcache support is an intentional non-goal, #119) runs `composer test:opcache-runner` plus the existing `test:opcache` non-skip gate. Docs: hot-swap support-matrix rows for both new behaviours, self-debugging optimizer caveat, opcache-binary Windows non-goal, AGENTS.md group rules.

- PHP version (full first line of `php -v`): PHP 8.4.19 (cli) (built: Mar 30 2026 19:28:35) (NTS)
- Thread safety: NTS
- OS / architecture: Linux x86-64 (Ubuntu)

Verified: full suite with `opcache.enable_cli=0` (503 tests OK), full suite with `opcache.enable_cli=1` via `test:opcache-runner` (501 tests OK), `--group opcache --fail-on-skipped` (31 tests OK, includes the new child probes), PHPStan level max clean, `cs:check` clean. The same opcache-on failure diff was confirmed on PHP 8.5.9 against `master` before the fix, so the cascade should port textually.

## Checklist

- [x] Targets the **minimum affected version branch** (`8.4`) — fixes cascade upward, never downward
- [x] `composer test` passes on the matching PHP minor
- [x] Tests added or updated; no new struct dereferences (the map-ptr read reuses `zend_op_array`/`HashTable`, already in `layout_structs`)
- [x] `tools/generator/symbols.php` unchanged — nothing under `include/`, `stubs/` or `.phpstorm.meta.php` touched

https://claude.ai/code/session_01BDcCQiYqbMkjRPyhWgLL6M

---
_Generated by [Claude Code](https://claude.ai/code/session_01BDcCQiYqbMkjRPyhWgLL6M)_


lisachenko:
The first run of the new opcache-active runner job earned its keep immediately: it failed on two tests that were green locally — `RedefineLeakPlateauTest` and `PersistentHeapRequestCycleTest`. Root cause: both child processes inherited `opcache.enable_cli` from php.ini, so their outcome depended on the environment (the job writes `opcache.enable_cli=1` into php.ini; Ubuntu's PHP 8.5 package even ships it `On` by default). The failures are real opcache bugs on **both** lines — the first `redefine()` not taking effect, and a registered module entry invisible to `extension_loaded()` — now tracked with reproductions in #242 and #243 (corrected from their original PHP-8.5.9 mis-attribution).

Pushed `3b3fc99`: the two children pin `-d opcache.enable_cli=0` (with `TODO(#242)`/`TODO(#243)` to unpin once fixed), making the leak-plateau and persistent-heap measurements hermetic everywhere — including Ubuntu-8.5 dev machines, where `composer test` on `master` was already failing for this reason.

---

PR #265, opened 2026-08-19
docs(opcache): ZCSG publication feasibility spike (#121) — NO-GO

### LINKED ISSUE #121: feat(opcache): load patched binaries into opcache shared memory (ZCSG)
Follow-up to the OpCache binary-file API (#111, merged). Explicitly deferred in `docs/opcache-binary.md`.

Today the patch→refresh loop is proven under `opcache.file_cache_only=1`; when opcache also uses shared memory, a resident script is only re-read after invalidation. The next step is to publish a patched image directly into the running accelerator's shared memory (ZCSG) so all workers pick it up without a file-cache round-trip.

**Scope (research + implementation):** bind the non-public opcache SHM entry points (`zend_file_cache_script_load` / `zend_accel_hash_update` / `ZCSG(hash)`), or an equivalent path, to install a patched `zend_persistent_script` into SHM safely across workers. This depends on `opcache.so` symbol visibility and careful locking; likely large and fragile across builds. Foundation for #121 (hot-swap wiring).

lisachenko:
## What this changes

Research deliverable for #121 (publish patched binaries directly into live opcache shared memory / ZCSG). **Verdict: NO-GO against stock opcache.** Docs-only — lands the evidence in-repo (`docs/research/zcsg-spike.md`) plus the four throwaway probe scripts under `tools/research/` (marked spike, never wired into `src/`). Closes #121 as not-planned.

### Load-bearing evidence

1. **Write-path symbols are unreachable.** FFI on Linux binds through `RTLD_DEFAULT`, so only default-visibility dynamic symbols resolve. `nm -D` on three builds (host release 8.4.19, container no-debug + debug 8.4.24) shows `opcache.so` exports only **6** defined dynamic symbols; every needed write-path symbol (`zend_shared_alloc`, `zend_accel_hash_update`, `zend_shared_alloc_lock`, `zend_accel_shared_protect`, `zend_file_cache_script_load`, the `accel_shared_globals`/ZCSG pointer, the lock fd) is hidden-visibility and **empirically fails** to bind via `FFI::cdef`, while the one `ZEND_EXT_API` symbol `smm_shared_globals` binds. No ini flag or z-engine change can export them.
2. **Read side IS reachable** — a prototype recovers the ZCSG base via `smm_shared_globals` and walks `ZCSG(hash)` live, reading resident script names from interned `zend_string` keys, with zero opcache symbols. The write sequence is blocked at the cross-process fcntl lock (unreachable anonymous fd) and safe interned-string mutation.
3. **Safety hazards are structural** — a userland SHM write succeeds under `opcache.protect_memory=0` but **SIGSEGVs under `=1`** (segment is mprotect'd read-only outside opcache's unprotect window); no lock, no safe interning, no crash atomicity — none mitigable against stock opcache.
4. **The alternative already ships** — `BinaryCacheFile::refresh()` drives opcache's own `zend_file_cache_script_load()` (the entire locked install) on a compile miss; FPM workers share one SHM segment, so one faulting request republishes for all. Direct writes would add only "push-update without a warm-up request," not worth racing every worker without opcache's lock. If that ever becomes a hard requirement, the honest route is a companion opcache/extension patch re-exporting the locked install path — not FFI, which defeats z-engine's stock-PHP premise.

- PHP version (full first line of `php -v`): PHP 8.4.19 (cli) (built: Mar 30 2026 19:28:35) (NTS); also container no-debug + debug 8.4.24
- Thread safety: NTS
- OS / architecture: Linux x86-64 (Ubuntu)
- Debug build (`--enable-debug`)? symbol inventory + protect-write crash reproduced across release, no-debug and debug builds

Docs-only change: PHPStan level max clean, cs-fixer clean.

## Checklist

- [x] Targets the **minimum affected version branch** (`8.4`)
- [x] `composer test` passes on the matching PHP minor (no test changes)
- [x] Tests added or updated — n/a (research/docs only)
- [x] `tools/generator/symbols.php` unchanged — nothing generated touched

https://claude.ai/code/session_01BDcCQiYqbMkjRPyhWgLL6M

---
_Generated by [Claude Code](https://claude.ai/code/session_01BDcCQiYqbMkjRPyhWgLL6M)_

### DIFF
--- a/docs/research/zcsg-spike.md
+++ b/docs/research/zcsg-spike.md
@@ -1,0 +1,313 @@
+# ZCSG publication feasibility spike (#121)
+
+**Question.** Can z-engine publish a patched cache-binary image *directly* into a
+live opcache shared-memory segment (ZCSG) — installing a `zend_persistent_script`
+into `ZCSG(hash)` so every worker serves it immediately — without going through a
+file-cache round trip?
+
+**Answer up front: NO-GO for direct ZCSG writes; the already-shipped
+file-cache→SHM reload path (PR #251/#253) is the sound solution and direct writes
+add almost nothing over it while carrying structural, unmitigable risk.** The
+control-plane entry points opcache would need (`zend_shared_alloc`,
+`zend_accel_hash_update`, `zend_shared_alloc_lock`, `zend_accel_shared_protect`,
+`zend_file_cache_script_load`) are **not exported** from `opcache.so` and are
+therefore unreachable through FFI on Linux. The data plane (the ZCSG pointer, the
+hash, the segment) *is* reachable and was walked live in a prototype, but reading
+it is not the hard part — writing it correctly across processes is, and that is
+exactly what the missing symbols provide.
+
+This is a research spike. All prototype code lives under `tools/research/` and is
+never wired into `src/`.
+
+---
+
+## 1. Symbol inventory (evidence)
+
+FFI on ELF binds engine symbols out of the **running process image via
+`RTLD_DEFAULT`** — `Core::engineLibrary()` returns `null` on `/`-separator hosts,
+so `FFI::cdef($defs, null)` resolves through the process-global dynamic symbol
+table. A symbol is bindable **iff** it appears in some loaded object's dynamic
+symbol table with **default visibility**. PHP builds every extension with hidden
+visibility and exports only symbols explicitly marked `ZEND_API` / `ZEND_EXT_API`.
+
+### `opcache.so` exports only 6 defined dynamic symbols — on every build
+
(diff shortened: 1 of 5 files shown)


lisachenko:
I believe we can just keep it as-is, unmerged, decision is recorded as a MD file

lisachenko:
Works for me — leaving it open and unmerged. The NO-GO decision and the load-bearing evidence are also captured in #121's closing comment, so it's discoverable from the epic without needing this branch merged; this PR just keeps the full report + reproducible probe scripts around if anyone wants to re-run the symbol inventory later. I'll stop tracking it.

---

PR #281, opened 2026-08-29
fix: user opcode handlers under PHP 8.6's tail-call VM (macOS arm64)

### LINKED ISSUE #280: User opcode handlers corrupt VM state under PHP 8.6's tail-call VM (macOS arm64)
## Summary

On macOS **arm64** with PHP **8.6** (Homebrew/setup-php build), installing a user opcode handler via `OpCode::setHandler()` on `EXT_STMT` (with `Compiler::COMPILE_EXTENDED_STMT` enabled) corrupts VM state as soon as the handler's PHP closure runs: closure bound variables read back as garbage (e.g. an `int` counter reads as a `ZEngine\Core` object), and full consumers (lisachenko/zdebug) segfault (`EXC_BAD_ACCESS at 0x10` inside `ZEND_DO_UCALL_SPEC_RETVAL_USED_TAILCALL_HANDLER`).

The same code is green on:
- macOS arm64 + PHP **8.5** (identical probe run as control — all steps pass)
- macOS **x64** (Intel) + PHP 8.6 — same PHP source build
- Linux x64 + PHP 8.6

z-engine's own CI is green on macOS arm64 + 8.6 because `OpCodeHookTest` hooks `ADD` inside a small probe function without `COMPILE_EXTENDED_STMT`; the corruption shows once EXT_STMT-instrumented code (compiled after the handler is installed) starts dispatching through the hook.

## Why only arm64 + 8.6

PHP 8.6 introduced a new VM dispatch kind, `ZEND_VM_KIND_TAILCALL` (`Zend/zend_vm_opcodes.h` in php-8.6.0beta2):

```c
#elif (defined(__GNUC__) && defined(HAVE_GCC_GLOBAL_REGS))
# define ZEND_VM_KIND  ZEND_VM_KIND_HYBRID
#elif defined(HAVE_MUSTTAIL) && defined(HAVE_PRESERVE_NONE) && (defined(__x86_64__) || defined(_M_X64) || defined(__aarch64__)) && defined(__clang__)
# define ZEND_VM_KIND  ZEND_VM_KIND_TAILCALL
#endif
```

- Linux (gcc) and macOS x64 (clang with global register support) build the **HYBRID** VM — unaffected.
- macOS arm64 clang has no `HAVE_GCC_GLOBAL_REGS`, so the build selects **TAILCALL**: opcode handlers become `const zend_op *(*)(zend_execute_data *, const zend_op *)` with the `preserve_none` calling convention, chained via `musttail`. The crash frames (`*_TAILCALL_HANDLER`) confirm the failing build runs this VM.

`zend_user_opcode_handlers[opline->opcode](execute_data)` is still called with the classic `int (*)(zend_execute_data *)` signature from `ZEND_USER_OPCODE_SPEC_TAILCALL_HANDLER`, so the trampoline installation itself works (a trivial handler on `ADD` fires fine). The breakage appears when the handler re-enters PHP execution (the FFI callback runs a userland closure) while the interrupted frame's opcode came from EXT_STMT-instrumented code — pointing at an interaction between libffi callback re-entry, nested `execute_ex`, and the `preserve_none` handler chain. It may ultimately be a php-src beta bug rather than z-engine's, but z-engine is where consumers hit it, and where a repro/workaround can live.

## Minimal repro (pure z-engine, no consumer code)

`probe.php`:
```php
<?php
require __DIR__ . '/vendor/autoload.php';

use ZEngine\Core;
use ZEngine\System\Compiler;
use ZEngine\System\OpCode;

Core::init();
Core::$compiler->setOptions(Core::$compiler->getOptions() | Compiler::COMPILE_EXTENDED_STMT);

$fires = 0;
OpCode::setHandler(OpCode::EXT_STMT, function ($scope) use (&$fires): int {
    $fires++;
    return Core::ZEND_USER_OPCODE_DISPATCH;
});

require __DIR__ . '/payload.php'; // any file: a class, new, a method call, a function call
echo "EXT_STMT fires: {$fires}\n";
```

Run with `php -d ffi.enable=1 -d opcache.jit=off probe.php` on macOS arm64 + PHP 8.6.0-dev:

```
PHP Warning:  Uncaught TypeError: Cannot increment ZEngine\Core in probe.php:12
PHP Fatal error:  Throwing from FFI callbacks is not allowed in payload.php on line 15
```

The by-ref `use (&$fires)` int reads back as a `ZEngine\Core` object on the very first handler invocation. Adding `RETURN`/`THROW` handlers fails the same way (`Cannot use object of type ZEngine\Core as array` on an array counter). On PHP 8.5 (same machine, same script) everything passes.

## Crash signature from a full consumer (zdebug)

```
exception: EXC_BAD_ACCESS / SIGSEGV, KERN_INVALID_ADDRESS at 0x0000000000000010
frame 0: ZEND_DO_UCALL_SPEC_RETVAL_USED_TAILCALL_HANDLER + 116
frame 1: execute_ex + 156
frame 2: zend_execute.cold.1 / zend_execute / zend_execute_script / php_execute_script_ex
```

(reading offset 0x10 of a NULL `zend_function`, i.e. a call frame whose `func` was clobbered). Downstream CI evidence: lisachenko/zdebug#24 — the diagnostic job `Diagnose arm64 (PHP 8.5/8.6)` in run https://github.com/lisachenko/zdebug/actions/runs/33254290174 shows the layered probes (L1 `Core::init` ✅, L2 `COMPILE_EXTENDED_STMT` ✅, L3 EXT_STMT handler ❌ on 8.6 / ✅ on 8.5, L5 module registration ✅, L6 full boot ❌ SIGSEGV) plus the lldb backtrace and the macOS `.ips` crash report.

## Suggested directions

- Reproduce with a C user opcode handler (no FFI) to split "TAILCALL VM vs user opcode handlers re-entering the VM" (a php-src bug worth reporting upstream while 8.6 is in beta) from "TAILCALL VM vs FFI callback trampolines" (z-engine-side).
- If it is php-src's: report before 8.6.0 GA; the tail-call VM is new in this cycle.
- Until resolved, consumers on Apple Silicon can be pointed at a HYBRID/CALL-VM PHP build; z-engine could also detect `ZEND_VM_KIND_TAILCALL` at `Core::init()` (e.g. via `php -i`/`Reflection` on the build metadata or a runtime probe) and fail fast with a clear message instead of corrupting the debuggee.

lisachenko:
## What

Fixes #280 on z-engine's side: on PHP 8.6 builds using the new **tail-call VM** (`ZEND_VM_KIND_TAILCALL` — clang without global-register support, notably Apple Silicon macOS), user opcode handlers mis-resume execution inside the engine and corrupt the process. The diagnostics on this branch reduced it to a **php-src bug** (pure ext-ffi repro, no z-engine code — see the [upstream-ready report on #280](https://github.com/lisachenko/z-engine/issues/280#issuecomment-5463036427)): the generated `ZEND_USER_OPCODE_SPEC_TAILCALL_HANDLER` returns the single-step dispatch result up the musttail chain and `execute_ex()` resumes it against its stale entry frame.

Until php-src resolves it, z-engine fails fast instead of corrupting the debuggee:

- **`Core::vmKind()`** — reports `zend_vm_kind()` through a dedicated one-symbol FFI binding (usable before `init()`, no generated-header changes), with `VM_KIND_*` constants mirroring `Zend/zend_vm_opcodes.h`
- **`OpCodeHook::install()`** throws `OpCodeHookException::tailCallVmUnsupported()` on `VM_KIND_TAILCALL`, naming the issue
- **`OpCodeHookVmKindGuardTest`** runs in every CI leg (deliberately **not** in the `internal` group — the existing opcode-hook tests are internal-only, which is why the macOS legs never caught this): on tail-call builds it asserts the refusal, elsewhere the unchanged install/uninstall lifecycle
- README/AGENTS document the platform caveat

## Diagnostics kept as the upstream repro harness

`tools/diagnostics/issue-280/` (probe ladder + `pure-ffi-repro.php`) and the now **dispatch-only** `Diagnose issue 280` workflow stay in-tree: re-dispatching it against a newer PHP build answers "is the php-src bug fixed yet"; both go away with the guard once upstream resolves it.

## Verification

- **macOS arm64 (tail-call VM)**: full CI green with the guard test taking the refusal branch; the pure-FFI repro crashes there (`vm_kind=5`, SIGSEGV) proving the underlying bug is engine-level — [run 33257997984](https://github.com/lisachenko/z-engine/actions/runs/33257997984)
- **macOS x64 / linux (hybrid VM)**: all probes and the repro pass; full suite, PHPStan level max, cs — green
- Probe evidence for the mechanism (stage-marked ladder, lldb, crash reports): [run 33256998594](https://github.com/lisachenko/z-engine/actions/runs/33256998594)

Downstream: lisachenko/zdebug#24 keeps its macOS arm64 + 8.6 leg experimental until upstream is fixed; with this guard, consumers get a clear `OpCodeHookException` instead of corrupted debuggees.

---

https://claude.ai/code/session_01U4cKi87mVQ896uSrw592fG

### DIFF
--- a/.github/workflows/diagnose-280.yml
+++ b/.github/workflows/diagnose-280.yml
@@ -60,6 +60,10 @@ jobs:
             php $PHP_FLAGS tools/diagnostics/issue-280/probe.php "$mode" || echo "exit=$?"
           done
 
+      - name: Pure-FFI repro (no z-engine, for the php-src report)
+        if: always()
+        run: php $PHP_FLAGS tools/diagnostics/issue-280/pure-ffi-repro.php || echo "exit=$?"
+
       - name: noop under lldb (backtrace on crash)
         if: always()
         run: |
--- a/.github/workflows/diagnose-280.yml
+++ b/.github/workflows/diagnose-280.yml
@@ -1,0 +1,96 @@
+# TEMPORARY: diagnostics for issue #280 (user opcode handlers corrupt VM state
+# under PHP 8.6's tail-call VM on macOS arm64). Runs the tools/diagnostics/issue-280
+# probe ladder on the failing runner (macos-latest, arm64) with macos-15-intel as
+# the in-run control. Removed together with the probes once the fix lands.
+name: Diagnose issue 280
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: diagnose-280-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  PHP_MINOR: '8.6'
+
+jobs:
+  probes:
+    name: Probes (${{ matrix.runner-arch.arch }})
+    runs-on: ${{ matrix.runner-arch.runner }}
+    env:
+      HOMEBREW_NO_AUTO_UPDATE: '1'
+      HOMEBREW_NO_INSTALL_CLEANUP: '1'
+      ZENGINE_STRICT_LAYOUT_CHECK: '1'
+      PHP_FLAGS: -d ffi.enable=1 -d zend.assertions=1 -d opcache.enable_cli=0 -d opcache.jit=off
+    strategy:
+      fail-fast: false
+      matrix:
+        runner-arch:
+          - { runner: macos-latest, arch: arm64 }
+          - { runner: macos-15-intel, arch: x64 }
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ env.PHP_MINOR }}
+          extensions: ffi, opcache
+          ini-values: ffi.enable=1, zend.assertions=1, opcache.enable=1, opcache.enable_cli=0, opcache.jit=off, opcache.jit_buffer_size=0
+          coverage: none
+
+      - name: Environment report
+        run: |
+          uname -m
+          php -v
+          php -r 'echo "ZEND_THREAD_SAFE=", var_export(ZEND_THREAD_SAFE, true), " PHP_DEBUG=", PHP_DEBUG, PHP_EOL;'
+
+      - name: Install dependencies
+        uses: ramsey/composer-install@v4
+
+      - name: 'Probe: noop (handler touches nothing)'
+        if: always()
+        run: php $PHP_FLAGS tools/diagnostics/issue-280/probe.php noop; echo "exit=$?"
+
+      - name: 'Probe: log-const (one internal call per fire)'
+        if: always()
+        run: php $PHP_FLAGS tools/diagnostics/issue-280/probe.php log-const; echo "exit=$?"
+
+      - name: 'Probe: globals (symbol-table counter)'
+        if: always()
+        run: php $PHP_FLAGS tools/diagnostics/issue-280/probe.php globals; echo "exit=$?"
+
+      - name: 'Probe: use-ref (by-ref closure counter, the failing shape)'
+        if: always()
+        run: php $PHP_FLAGS tools/diagnostics/issue-280/probe.php use-ref; echo "exit=$?"
+
+      - name: 'Probe: diag (engine invariants per fire)'
+        if: always()
+        run: php $PHP_FLAGS tools/diagnostics/issue-280/probe.php diag; echo "exit=$?"
+
+      - name: 'Probe: add-baseline (ADD handler, no extended stmt)'
+        if: always()
+        run: php $PHP_FLAGS tools/diagnostics/issue-280/probe.php add-baseline; echo "exit=$?"
+
+      - name: 'use-ref under lldb (backtrace on crash)'
+        if: always()
+        run: |
+          lldb --batch \
+            -o 'run' \
+            -k 'thread backtrace all' \
+            -k 'quit' \
+            -- "$(which php)" $PHP_FLAGS tools/diagnostics/issue-280/probe.php use-ref || true
+
+      - name: macOS crash reports
+        if: always()
+        run: |
+          for f in ~/Library/Logs/DiagnosticReports/php*; do
+            [ -e "$f" ] || continue
+            echo "===== $f ====="
+            head -c 6000 "$f"
+          done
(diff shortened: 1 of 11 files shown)


lisachenko:
@claude Look for the relevant bug report for PHP at https://github.com/php/php-src/ (issues + PRs) and link it here if you will find anything relevant

lisachenko:
Searched php/php-src issues and PRs (queries: `zend_set_user_opcode_handler`, `ZEND_USER_OPCODE_DISPATCH`, `USER_OPCODE`, "user opcode handler" + tailcall/tail-call, plus web search) — **no existing report covers this bug**. Nobody has filed the user-opcode-handler mis-resume under the tail-call VM yet, so the report drafted on #280 is still worth submitting upstream.

Closest related upstream items (same bug class — the tail-call VM returning a next-opline/tagged value that the caller mishandles — but all about `vm_interrupt`, not user opcode handlers):

- [php/php-src#17849](https://github.com/php/php-src/pull/17849) — "Tail call VM" by arnaud-lb: the PR that introduced the tail-call VM (handlers return the next opline up the musttail chain), useful context and the right people to CC (arnaud-lb, morrisonlevi).
- [php/php-src#21922](https://github.com/php/php-src/pull/21922) — "Fix tailcall helper dispatch after VM interrupt" (merged, PHP-8.5): `ZEND_VM_DISPATCH_TO_HELPER()` dereferenced an opline tagged with `ZEND_VM_ENTER_BIT` returned by a helper during an interrupt — precisely the same *tagged/next-opline mishandled by the resuming code* pattern our diagnosis found in `ZEND_USER_OPCODE_SPEC_TAILCALL_HANDLER`'s `ZEND_USER_OPCODE_DISPATCH` path.
- [php/php-src#22265](https://github.com/php/php-src/pull/22265) — "Fix another vm_interrupt bug for tailcall VM" (merged, PHP-8.5): a follow-up crash in the same area (missing `SAVE_OPLINE` before `ZEND_VM_LOOP_INTERRUPT`).

Those two merged fixes show the pattern is a known weak spot, which should make the upstream report easy to triage — it can reference them, the pure-FFI repro (`tools/diagnostics/issue-280/pure-ffi-repro.php`, no z-engine code involved), and the analysis in [the #280 comment](https://github.com/lisachenko/z-engine/issues/280#issuecomment-5463036427). I can't file it from this session (no access to php/php-src), so that comment is ready to copy-paste as the issue body.

---

